PR #8, opened 2021-07-22
How to build firmware README.md

adigyran:
Quick guide how to build and upload firmware


adigyran:
it might be better to put in the wiki with all troubleshooting

PR #18, opened 2021-08-22
Using common I2C pins for ESP32 to prevent lockup

JINODK:
Using common I2C pins for ESP32 to prevent lockup when using MPU
Pin 12 cause boot to fail when pull to HIGH (which it does on mpu6050)
image
image

Source: https://randomnerdtutorials.com/esp32-pinout-reference-gpios
Pinout: https://lastminuteengineers.com/esp32-arduino-ide-tutorial/#esp32-development-board-pinout


Eirenliel:
Thanks. Also removed this pin from I2C discovery

PR #22, opened 2021-09-02
Fix to store credentials for ESP8266

### LINKED ISSUE #20: Wifi credentials doesn't save after reset
Wifi credentials set by serial command don't recover after reset. Have to set each time (and also impossible to use because at the same moment i disconnect usb, it already gone)
Only hardcode credentials work fine

Tested on 2 node mcu v2

Kamilake:
Fixes #20 

I have modified the code to be ESP8266 specific so I can store credentials in EEPROM(using DeviceConfig).
I tested it on ESP32(WROOM32, WROVER-B) and 8266(nodemcu ESP12F) with an MPU9250 and everything seems to be working fine.

### DIFF
--- a/src/configuration.h
+++ b/src/configuration.h
@@ -39,6 +39,8 @@ struct DeviceConfig {
     CalibrationConfig calibration;
     int deviceId;
     int deviceMode;
+    char wifissid[50];
+    char wifipass[50];
 };
 
 DeviceConfig * const getConfigPtr();
--- a/src/wifihandler.cpp
+++ b/src/wifihandler.cpp
@@ -45,12 +45,22 @@ bool isWiFiConnected() {
 
 void setWiFiCredentials(const char * SSID, const char * pass) {
     WiFi.stopSmartConfig();
+#ifndef ESP32
+    DeviceConfig * const config = getConfigPtr();
+    memcpy(config->wifissid, SSID, strlen(SSID));
+    memcpy(config->wifipass, pass, strlen(pass));
+    config->wifissid[strlen(SSID)] = 0; //EOL
+    config->wifipass[strlen(pass)] = 0;
+    saveConfig();
+#endif
     WiFi.begin(SSID, pass);
     wifiState = 2;
     wifiConnectionTimeout = millis();
 }
 
 void setUpWiFi() {
+    DeviceConfig * const config = getConfigPtr();
+    
     Serial.println("[NOTICE] WiFi: Setting up WiFi");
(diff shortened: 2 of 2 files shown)


Eirenliel:
I don't understand why it saves credentials for me 100% of the time, but doesn't save for many others? 8266 wifi library also has eeprom wifi credentials storage, this shouldn't be required. That's why I didn't just save it in eeprom myself, I want to first investigate why it does or doesn't work in different cases/chips.

Eirenliel:
According to [docs](https://arduino-esp8266.readthedocs.io/en/latest/esp8266wifi/station-class.html#begin) this should work: https://github.com/SlimeVR/SlimeVR-Tracker-ESP/commit/65140fa445fe3932ba312c3aafcf83ade9f3f141

adigyran:
> According to [docs](https://arduino-esp8266.readthedocs.io/en/latest/esp8266wifi/station-class.html#begin) this should work: [65140fa](https://github.com/SlimeVR/SlimeVR-Tracker-ESP/commit/65140fa445fe3932ba312c3aafcf83ade9f3f141)

4 boards DOIT nodemcu v2 tested, this fix is not working

Kamilake:
Thank you for your reply. I'm curious about the same part.

https://arduino-esp8266.readthedocs.io/en/latest/esp8266wifi/station-class.html#begin

Obviously according to that documentation Wifi.begin() should work.
Actually that code works without problems on ESP32 but not on ESP8266... and I couldn't find an example of using that code.
I don't think this is a fault with my board. I think there is something wrong with the ESP library or code.
Does anyone know why the ESP8266 can't store credentials?

Eirenliel:
Are you in the discord @Kamilake? We were discussing it there too, seems like it's not working for most, but is working for some.

Kamilake:
> Are you in the discord @Kamilake? We were discussing it there too, seems like it's not working for most, but is working for some.

I don't use discord but I'm willing to go in there if there's a discussion.
Can I go there? I'd like to check the case of working well..  
I'll be there after work today. @Eirenliel

Eirenliel:
Don't need if you don't want to :D Your fix should work fine, but I am hesitant to just merge it without figuring out why it doesn't work on its own. It does work for my prototype slimes, and it works for some other users that built DIY slimes.

adigyran:
Latest changes works fine

adigyran:
still have #23

Eirenliel:
Seems like persistent true fixed the problem.

PR #29, opened 2021-09-27
Support for dual MPU-6050's

carl-anders:
This pull request adds support for using two MPU-6050's connected to one device.

This mode is more subject to interference than running in single MPU mode. Good solder joints, cables and a good Wi-Fi connection is needed to keep the TPS high. I can run 4 normal trackers, and one with dual on my Wi-Fi with minimal issues, but adding more dual trackers unfortunately makes my router sad.

Any future documentation for this mode should explain these issues (I will simply create a discord post for now). If the user has problems with TPS they should try to unplug the second tracker and run it in single MPU mode.

### DIFF
--- a/src/defines.h
+++ b/src/defines.h
@@ -63,6 +63,8 @@
   #define IMU_HAS_MAG false
   #define I2C_SPEED 100000
   #define IMU_MPU6050_RUNTIME_CALIBRATION // Comment to revert to startup/traditional-calibration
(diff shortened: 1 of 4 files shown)


Eirenliel:
Can this be improved by utilizing and int pin instead of polling i2c?

carl-anders:
There's a good possibility yes, and it's definitely something to test out. Using the more lenient polling code made by masterex1000 would probably also improve things. But I have spent many hours in vrchat with only these changes and no major issues after I turned off some other Wi-Fi devices clogging my network.

PR #32, opened 2021-10-24
Calibration with EEPROM for MPU9250

LETS-BEE:
when we use MPU9250, server show NaN data. Because calibration data is not in EEPROM. So, Add Calibration Codes.
I refer to megneto 1.4 version(https://github.com/jremington/MPU-9250-AHRS)
and I think you refer to jremington's codes for MPU9250.

Change mpu9250Sensors.cpp
Change sensors.h
Change configuration.h
Add megneto1.4.h


Eirenliel:
I will test, refactor and merge it when I have time, thanks for your work 👍

LETS-BEE:
Better codes in #55

PR #40, opened 2021-11-22
Make 400khz speed default for MPU sensors 

adigyran:
People tested and found that 400khz is more stable for MPUs also


Eirenliel:
Isn't either 6050 or 6500 not support 400kHz?

carl-anders:
Both support 400kHz according to datasheets.
I only have 6050's myself, but they do work at 400kHz, and it's required to run dual MPU's.

PR #42, opened 2021-11-27
ESP-01 support + VCC Monitor

Levi--G:
ESP-01 support:
-Added defines
-Moved led operations to ledmgr
-Added option to disable LEDS
VCC monitor support:
-Added VCC monitor, not compatible with using internal ADC
-Every ESP has a slight offset to the actual voltage, but relative voltage is stable and usable
-Preliminary configuration will warn low battery at 3.2V and shut down at 3.1V


adigyran (approved):
Great work

adigyran:
oh i have ESP-01 gonna see how it works

Eirenliel:
why adding new led manager instead of expanding led status?

Levi--G:
> why adding new led manager instead of expanding led status?

To be honest, i didn't quite get the purpose of the status class, it wasn't being used in the imu drivers or where the other leds were set in main etc and i don't know how i would add the calls to that class. The ledstatus also just handles the status led, while the other classes use the 2 other led numbers. But if you tell me how you want it added i can move it? :)

Kamilake:
that seems to be very effective for diy users!

Levi--G:
Just revised the battery system on my fork, moved the logic to a sepperate class and simplified code where available. Made slight adjustment to the defines to be a bit cleaner and added some explanation what does what. Also added compiler warnings for incompatible config.

Eirenliel:
Pushed it in test branch, will do my own refactoring and some other changes before merging into main

PR #43, opened 2021-11-29
MPU6050 MotionApps bugfix

Levi--G:
Since the IMU_MPU6050_RUNTIME_CALIBRATION is checked before sensor.h is included (and thus before defines.h/debug.h) it is in fact at that point not set yet.
So basically the old MotionApps20 is always included. If we move defines.h before motionapps that should be fixed.

### DIFF
--- a/src/mpu6050sensor.cpp
+++ b/src/mpu6050sensor.cpp
@@ -21,6 +21,8 @@
     THE SOFTWARE.
 */
 
+#include "defines.h"
+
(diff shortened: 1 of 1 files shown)


carl-anders:
Change tested with Wemos D1 Mini + 2 x MPU6050. Still works. Unknown if any improvement in tracking, but before it did indeed only load "20" version, not "v6_12". This means that the "IMU_MPU6050_RUNTIME_CALIBRATION" def actually does something.

PR #48, opened 2021-12-06
Add support for Mixing BNO08X and MPU6XXX IMU's

carl-anders:
This code is currently untested for any configurations except:
* MPU6050 in following configurations
* * Two of them
* * One in 0x68
* * One in 0x69

**This pull request needs to be tested** to make sure 2xBNO085 works, and mixing works.

If merged I will also fix the docs to include the ability to do BNO + MPU. (It will also add `IMU_MPU6050_RUNTIME_CALIBRATION`)

**Future work:**

* Needs more work in defines.h. Needs to add info on how to mix and match.
* We need the ability to mix and match other things, but current defines makes things weird. this commit only works as we don't use has_accell, has_gyro, has_mag. Need to do solve issue for `BNO_HAS_ARVR_STABILIZATION` to mix and match  BNO's.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -35,21 +35,42 @@
 
 #if IMU == IMU_BNO080 || IMU == IMU_BNO085
     BNO080Sensor sensor{};
-    #if defined(PIN_IMU_INT_2)
-        #define HAS_SECOND_IMU true
-        BNO080Sensor sensor2{};
-    #endif
 #elif IMU == IMU_BNO055
     BNO055Sensor sensor{};
 #elif IMU == IMU_MPU9250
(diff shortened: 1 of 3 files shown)


Louka3000:
> This pull request needs to be tested to make sure 2xBNO085 works, and mixing works.

^^ Tested with 2 BNO085 and 1 BNO085 and it worked in both cases :3

MightyGood:
Tested with BNO085 as primary and MPU-9250 (defined as MPU-6500) and it worked. Each IMU was a tracker that moved independently.

Eirenliel:
Included in #50

PR #49, opened 2021-12-08
Replace the WiFiUDP library with the AsyncUDP library 

Sark1tama:
Tested available at ESP32, Ping min 1-3ms, average less than 10ms, and stable. When using SEND_UPDATES_UNCONNECTED may cause packet reception failure before handshake is completed.


Eirenliel:
Is there any reason why this is better than what was before?

Sark1tama:
> Is there any reason why this is better than what was before?

The AsyncUDP library uses FreeRTOS to make UDP sending and receiving asynchronous

Eirenliel:
> > Is there any reason why this is better than what was before?
> 
> The AsyncUDP library uses FreeRTOS to make UDP sending and receiving asynchronous

Ah, it's for ESP32, and for ESP8266 it uses ESPAsyncUDP to have the same interface. This would require a lot of testing for ESP8266 before it can be merged, because ESP8266 is the main target platform.

Eirenliel:
Can you replace the library with https://github.com/OttoWinter/ESPAsyncTCP ? It's the fork used in Esphome, i trust it to work more as it looks maintainted better.

Eirenliel:
Forget the comment I just posted and removed :P bad reading comprehension, sorry!

PR #53, opened 2021-12-17
Improve magnetometer support for MPU-9250

deiteris:
> NOTE: This is not tested. Similar code was used on different magnetometer and it works. Any volunteers with MPU-9250 are welcome.

Improves magnetometer support for MPU-9250. This includes faster update rate (100hz), proper scaling using factory adjustments and mode correction, and possibility to run two 9250 without conflicts due to magnetometer.

This will require to add scaling factors to calibration process introduced in https://github.com/SlimeVR/SlimeVR-Tracker-ESP/pull/32.

### DIFF
--- a/lib/math/helper_3dmath.cpp
+++ b/lib/math/helper_3dmath.cpp
@@ -23,15 +23,18 @@
 
 #include "helper_3dmath.h"
 
-float vector_dot(float a[3], float b[3])
-{
-    return a[0] * b[0] + a[1] * b[1] + a[2] * b[2];
+// Fast inverse square root function (https://en.wikipedia.org/wiki/Fast_inverse_square_root).
+// Source: https://pizer.wordpress.com/2008/10/12/fast-inverse-square-root/
+float invSqrt(float x) {
+   uint32_t i = 0x5F1F1412 - (*(uint32_t*)&x >> 1);
+   float tmp = *(float*)&i;
+   return tmp * (1.69000231f - 0.714158168f * x * tmp * tmp);
 }
 
 void vector_normalize(float a[3])
 {
-    float mag = sqrt(vector_dot(a, a));
-    a[0] /= mag;
-    a[1] /= mag;
-    a[2] /= mag;
+    float norm = invSqrt(a[0] * a[0] + a[1] * a[1] + a[2] * a[2]);
+    a[0] *= norm;
+    a[1] *= norm;
+    a[2] *= norm;
 }
--- a/lib/math/helper_3dmath.h
+++ b/lib/math/helper_3dmath.h
@@ -215,8 +215,7 @@ class VectorFloat {
         }
 };
 
-float vector_dot(float a[3], float b[3]);
-
 void vector_normalize(float a[3]);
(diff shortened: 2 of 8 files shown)


Boronare:
Pull request after check if code working well please...

1.Build error cuz you deleted "CalibrationConfig * calibration;" so cannot find CalibrationConfig.
2.getMotion9 not working.

deiteris:
>1.Build error cuz you deleted CalibrationConfig * calibration; from mpu9250sensor.cpp at acd7b1a so cannot find CalibrationConfig.

Yeah, I missed that, sorry.

>2.getMotion9 not working.

Could you tell how exactly it doesn't work? Does it return all zeroes for mag or what? I don't have AK8963 so it's a bit hard to tell for me. I have QMC5883L and it works with similar code.

Boronare:
Yes. It returns all zeroes.
I looked for "reading mag data from slave device with master device(mpu6050)s register" and wrote some writebyte codes before reading, but still not worked...
I couldn't find the reason or how to fix it...

Are there codes for QMC working in your repository?

deiteris:
> I looked for "reading mag data from slave device with master device(mpu6050)s register" and wrote some writebyte codes before reading, but still not worked...

You can find similar code for this in kriswiner's repo: https://github.com/kriswiner/MPU9250/blob/master/AK8963_as_slave/MPU9250.cpp#L237

> Are there codes for QMC working in your repository?

I've uploaded the same changes to separate branch with adjustments removed (since QMC already outputs scaled data) and initializing function modified for it. Here you can see how it works:
image

Here's the code:
https://github.com/deiteris/SlimeVR-Tracker-ESP/blob/qmc-mag/lib/mpu9250/MPU9250.cpp#L56

Boronare:
I'm terribly sorry.
I ordered MPU6050+QMC5883L and checked your code working.
Then I tested your code again,
...It worked...

What I found is : by default, ext_sens_data is disabled when dmp enabled...
and then... I realized I didn't test plain version of your code... what a dumbass...
All my fault. Sorry again.


So... nice work. It seems much better approach than what I did.
Both for TPS and dual-MPU support.

deiteris:
>What I found is : by default, ext_sens_data is disabled when dmp enabled...
>and then... I realized I didn't test plain version of your code... what a dumbass...

Ah, I was wondering whether you cloned only these changes or merged them into yours, but had no idea.

If you want to use magnetometer with DMP, you can set up reading from slave sensor during DMP configuration, just before you reset DMP and FIFO. Otherwise, it won't work. Here's what I did in order to make it work with your code (using QMC5883L and same getMotion9 function as in this PR):
image

I didn't investigate how to add readings from slave sensor to FIFO packets (never checked where it is in the packet and we don't rely on it anyway), but it is possible if necessary.

>All my fault. Sorry again.

It's fine, no worries. Some things are confusing with registers and there isn't much knowledge unfortunately.

Boronare:
Thanks for information. I've solved compatibility problem.

Eirenliel:
Integrated into #55

PR #59, opened 2021-12-23
Add Wireless hotspot mode

loucass003:
I added a wireless hotspot mode. So you can change your wifi settings if you moved to another place or whatever.

I know there is already smartconfig, but i never managed to make it work so ......
Anyway the code for smart config is still here, now you can just choose between hotspot mode and smartconfig


src/wifihandler.cpp
```diff
@@ -72,6 +80,30 @@ void onConnected() {
     onWiFiConnected();
 }
 
+
+void wifiNoCredsloop() {
+    if (WIFI_NOCREDS_CONNECT == WIFI_SMARTCONFIG) {
```
deiteris, line 85: Since `WIFI_NOCREDS_CONNECT ` is a define, maybe convert this if operator to `#if/elif/endif` directive so it compiles only necessary code?
```c
#if WIFI_NOCREDS_CONNECT == WIFI_SMARTCONFIG
...
#elif WIFI_NOCREDS_CONNECT == WIFI_HOTSPOT
...
#endif
```

loucass003: i agree, i making the fix

loucass003:
Nobody wants to look at it ? i know it cannot be the final solution but its definatly better than nothing rn

Eirenliel:
I will be integrating it after I got all my stuff with the alghoritm I described in discord

Kamilake:
I think this is a very good function and necessary. In addition to being able to remotely control firmware, it also can use SlimeVR to be supported by low-cost ESP8266 modules (e.g., ESP-01) without USB ports or CH340 chipsets.

Eirenliel:
Please update it to the latest firmware changes, provisioning is now moved to wifiprovisioning.cpp file

Eirenliel:
[Add wireless hot-spot mode](https://trello.com/c/tLBBkXmM/13-add-wireless-hot-spot-mode)

Eirenliel (changes requested):
Please reabase against main, @loucass003

PR #60, opened 2021-12-23
Re-enable -O2 flag

deiteris:
My test with ESP32 and MPU-6050 didn't show any problems and the caused issue seems to be unrelated to this change.
https://discord.com/channels/817184208525983775/818062281636053012/922144740013649950

So let's try again maybe or conduct more tests?🤔


Eirenliel:
I guess we can try again :D

PR #70, opened 2022-01-05
Set baudrate to 74880 for esp8266

deiteris:
This fixes garbage characters during boot at a cost of lower serial speed in general. Helps with debugging weird board issues by looking at the boot mode:
https://riktronics.wordpress.com/2017/10/02/esp8266-error-messages-and-exceptions-explained/
image

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -11,7 +11,6 @@
 [env]
 lib_deps=
   https://github.com/pvizeli/CmdParser.git
-monitor_speed = 115200
 framework = arduino
 ;If you want to set hardcoded WiFi SSID and password, uncomment and edit the lines below
 ; '" - quotes are necessary!
@@ -30,11 +29,13 @@ framework = arduino
 ; Settings for different boards
 
 [env:esp12e]
+monitor_speed = 74880
 platform = espressif8266
 board = esp12e
 
 ; Uncomment below if you want to build for esp32
 ; Check your board name at https://docs.platformio.org/en/latest/platforms/espressif32.html#boards
 ;[env:esp32]
+;monitor_speed = 115200
 ;platform = espressif32
 ;board = esp32dev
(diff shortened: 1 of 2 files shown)


deiteris:
This also requires a corresponding change to baud rate on the server side, otherwise it's all garbage in WiFi window :\

Eirenliel:
I understand the idea, but this will make it incompatible with old firmware.. but probably now is the best time to do it.

Eirenliel:
Ah, I see you set different baud rates for different controllers. This is bad, because server can't know which one is connected before it is, it's not user-friendly...

deiteris:
Yeah, I realized this a bit later. Further idea was to detect baud rate based on the firmware and even gracefully handle compatibility with older firmware. Then I realized we don't report such data over serial...

The problem is that ESP32 uses 115200 baud for boot sequence, so if we keep 74800 baud for ESP8266, this will result in garbage boot log on ESP32. So we defo need a proper method to detect compatibility.

Eirenliel:
I understand the problem, I don't like the solution though.

Kamilake:
Is there any way to change its boot log to 115200?

deiteris:
> Is there any way to change its boot log to 115200?

As far as I understand, no, this is controlled by bootloader that sets the baud rate depending on the crystal used.
https://stackoverflow.com/a/41011637

So this makes it a hardware issue that cannot be easily resolved by the software.

> I understand the problem, I don't like the solution though.

I mean, if someday we have some proper protocol over serial, this is the only user-friendly option because no manipulations would be required from the user. Otherwise, you could add baud rate selection for ESP8266 and ESP32 in the server, which is less convenient, but an option too.

I'll close this PR since this is not as easy problem as I was imagining it, but it is something to keep in mind since it would simplify troubleshooting.

Kamilake:
> > Is there any way to change its boot log to 115200?
> 
> As far as I understand, no, this is controlled by bootloader that sets the baud rate depending on the crystal used. https://stackoverflow.com/a/41011637
> 
> So this makes it a hardware issue that cannot be easily resolved by the software.

Oh no... That was bad news. Why did Espressuf use unusual speeds? It never seems easy.

Thank you for the information.

PR #72, opened 2022-01-21
Adds support for ICM-20948

ThreadOfFate:
Adds support for ICM-20948 using's SparkFun's MIT library.
Currently supports 6DOF and 9DOF as options.


Eirenliel:
Please remove visual studio files from the PR and change the base branch to test - it's what will be soon merged ino main. Thans. Very excited for this :3

ThreadOfFate:
I have pushed a commit without the Visual Studio files, it seems to take some time to update the pull request.

src/icm20948sensor.cpp
```diff
@@ -0,0 +1,769 @@
...
+        working = false;
+        lastData = millis();        
+        Serial.print("[ERR] Sensor timeout ");
+        Serial.println(addr);
+    }
+}
+
+void ICM20948Sensor::sendData() { 
+    ICM_20948_Status_e readStatus = imu.readDMPdataFromFIFO(&dmpData);
+        if(readStatus == ICM_20948_Stat_Ok || ICM_20948_Stat_FIFOMoreDataAvail)
```
deiteris, line 459: This condition will not work as intended. You also need to check whether `readStatus == ICM_20948_Stat_FIFOMoreDataAvail`, otherwise right side is just always true.

PR #50, opened 2021-12-14
Refactor of sensors

Vyolex:
Refactor of the sensors code, tried to keep the substantial changes to a minimum to avoid breaking stuff. Nonetheless, a lot of code has changed and should be tested properly, only managed to test with the 6050s I have.

**Added:**
- `SensorFactory` class responsible for the creation of the sensors
- Ability to add second BNO055 (untested)
- Ability to mix different IMU types (code from #48 by @carl-anders), refactored and extended with ability to have BNO's as secondary IMU (untested)

**Changed:**
- Refactored sensor classes
- Fixed some bugs
- Accidentally auto formatted, sorry

Hopefully this helps a bit and thanks Moof to allow me to re-implement your code ❤️ 
Future work mentioned on their PR still applies.

PR #73, opened 2022-01-23
Add BNO055 Auxiliary Support

nerdywoffy:
It would be nice if BNO055 got aux tracker support too since it doesn't much different than BNO080 🤔


Eirenliel:
Thanks for the PR. Auxiliary support for all sensors was implemented in #50 and is currently in test branch. It will be merged in main either today or next week.

nerdywoffy:
Thanks for letting me know about it! can't wait to test the aux support for BNO055 later 🥳

PR #82, opened 2022-02-01
MPU9250: Switch to MotionApps V6.12

LETS-BEE:
Oringinal MotionApps is v4.1
I modify some code in MPU6050's MotionApps V6.12.
Work DMP, raw mag both.
and nomerlize in filter(for all sensor), add madwick filter.


lib/mpu9250/MPU9250.cpp
```diff
@@ -110,16 +120,12 @@ void MPU9250::getMagnetometerAdjustments(float adjustments[3]) {
...
-    return devAddr;
-}
-
 /** Verify the I2C connection.
  * Make sure the device is connected and responds as expected.
  * @return True if connection is valid, false otherwise
  */
 bool MPU9250::testConnection() {
-    return getDeviceID() > 0;
+    return getDeviceID() == 0x71;
```
deiteris: There's also MPU-9255 that has 0x73 device ID.
https://stanford.edu/class/ee267/misc/MPU-9255-Register-Map.pdf

LETS-BEE: Thank you!

LETS-BEE:
add float q[4]; param for raw filter(mahony filter)

PR #90, opened 2022-02-05
Move sendSignalStrength to wifihandler

unlogisch04:


Eirenliel:
oh, good catch! i was planning to do this during refactoring, but forgot, lol

PR #107, opened 2022-02-22
Fix compilation for ESP32

TheDevMinerTV:
Not the cleanest solution, but it works for now.
We need a equivalent method for this on the ESP32s.

### DIFF
--- a/src/serial/serialcommands.cpp
+++ b/src/serial/serialcommands.cpp
@@ -77,7 +77,7 @@ namespace SerialCommands {
             EEPROM.write(i, 0xFF);
         EEPROM.commit();
         WiFi.disconnect(true); // Clear WiFi credentials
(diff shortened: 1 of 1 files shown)


src/serial/serialcommands.cpp
```diff
@@ -77,7 +77,11 @@ namespace SerialCommands {
             EEPROM.write(i, 0xFF);
         EEPROM.commit();
         WiFi.disconnect(true); // Clear WiFi credentials
+        #if BOARD == BOARD_SLIMEVR || BOARD == BOARD_WEMOSD1MINI || BOARD == BOARD_NODEMCU || BOARD == BOARD_ESP01 || BOARD == BOARD_TTGO_TBASE
```
Eirenliel, line 80:
```suggestion
        #if BOARD == BOARD_SLIMEVR || BOARD == BOARD_WEMOSD1MINI || BOARD == BOARD_NODEMCU || BOARD == BOARD_ESP01 || BOARD == BOARD_TTGO_TBASE
        #if ESP8266
```

PR #115, opened 2022-03-01
Improve CI

TheDevMinerTV:
With this, we're now doing 2x8 (ESP8266 / ESP32 + the different IMUs) matrix builds, resulting in 16 builds.
Maybe we can split these up into parallel builds?


Eirenliel:
wouldn't this quickly consume our monthly actions quota?

PR #116, opened 2022-03-01
Faster upload speed for ESP8266

JINODK:
### Change the upload baud rate to 921600 (as CH340G can handle it) and save quite a lot of time actually

Before
`Wrote 389152 bytes (280698 compressed) at 0x00000000 in 24.8 seconds (effective 125.7 kbit/s)...`

After
`Wrote 389152 bytes (280698 compressed) at 0x00000000 in 4.1 seconds (effective 760.6 kbit/s)...`

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -40,6 +40,8 @@ build_unflags = -Os
 [env:esp12e]
 platform = espressif8266
 board = esp12e
+; Comment this line below if you have any trouble uploading the firmware
+upload_speed = 921600
 
 ; Uncomment below if you want to build for ESP-01
 ;[env:esp01_1m]


Kamilake:
Yes, that works fine with CH340, CP210x and most USB bridges.

 But... what I'm worried about is that some Windows drivers (Linux is fine) for the FT2232H and CP2102 may not support high-speed communication. 
(This comment is not an objection because everything worked for me!)

deiteris:
CP2102 is a part of CP210x family. Can confirm this works well with CP2102 on both NodeMCU ESP32 and Wemos D1 Mini (modification with CP2102). FTDI chips should also work well, but don't have any to test.

Kamilake:
> CP2102 is a part of CP210x family. Can confirm this works well with CP2102 on both NodeMCU ESP32 and Wemos D1 Mini (modification with CP2102). FTDI chips should also work well, but don't have any to test.

I'm always uploading at 921600 with cp2102 ch340, so I agree to increase the baudrate if there are no further issues.

JINODK:
> But... what I'm worried about is that some Windows drivers (Linux is fine) for the FT2232H and CP2102 may not support high-speed communication. (This comment is not an objection because everything worked for me!)

The CP2102 is a beast as it can do 3000000 (3mil) baud no problem (I tried it once on my only esp32 with cp2102)

Eirenliel:
should probably set it for esp32 too

Eirenliel:
the esp01 part is broken

JINODK:
> the esp01 part is broken

Yup I saw that

PR #118, opened 2022-03-05
Fix compilation on linux

ColdIce1605:
Fix Linux compilation.
Libraries are fun


TheDevMinerTV:
Not needed as I've already fixed this in 23b36d1 ^^

ColdIce1605:
Well thanks I missed that

PR #122, opened 2022-03-06
Rewrite the SensorFactory

TheDevMinerTV:


Eirenliel (changes requested):
Good stuff but needs changes ^^

TheDevMinerTV:
ℹ️ Ready for QA (again)

TheDevMinerTV:
ℹ️ Ready for review (again)

src/globals.h
```diff
@@ -29,7 +29,7 @@
 #include "defines.h"
 
 #ifndef SECOND_IMU
-#define SECOND_IMU IMU
+#define SECOND_IMU IMU_NONE
```
Eirenliel, line 32: I don't think it should default to NONE.

src/consts.h
```diff
@@ -24,6 +24,7 @@
 #define SLIMEVR_CONSTS_H_
 
 // List of constants used in other places
+#define IMU_NONE 0
```
Eirenliel, line 27: I don't think NONE is needed.

Eirenliel (changes requested):

src/sensors/ErroneousSensor.cpp
```diff
@@ -0,0 +1,48 @@
...
+        }
+
+        void ErroneousSensor::sendData()
+        {
+            if (m_SentFirstData)
+            {
+                return;
+            }
+
+            Network::sendRotationData(&quaternion, DATA_TYPE_NORMAL, 255, sensorId);
```
Eirenliel, line 43: This is not what it needs to send te be displayed, it should just return `SensorStatus::SENSOR_ERROR` in `getSensorState()`, everything else will be handled.

Eirenliel (changes requested):

unlogisch04 (changes requested):
Correct IMU debug text, and move motionLoop to a other position

src/sensors/SensorManager.cpp
```diff
@@ -0,0 +1,139 @@
...
+#elif IMU == IMU_ICM20948
+                    m_Sensor1 = new ICM20948Sensor(0, firstIMUAddress, IMU_ROTATION);
+#endif
+                }
+
+                m_Sensor1->motionSetup();
+            }
+
+            {
+#if SECOND_IMU == IMU_BNO080 || SECOND_IMU == IMU_BNO085 || SECOND_IMU == IMU_BNO086
```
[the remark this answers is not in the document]
TheDevMinerTV: Can't happen, this is solved by the redefine in `src/globals.h`

PR #126, opened 2022-03-12
Use POWER_SAVING_LEGACY by default

deiteris:
BNO users experience strange delays with new power saving options. Revert to legacy to mitigate the issue and for further investigation.


deiteris:
Closing my pull request in favor of #125.

PR #131, opened 2022-03-17
Update defines.h

rammlebubble:


TheDevMinerTV (changes requested):
Please fix your typos :)

rammlebubble:
> Please fix your typos :)

changed some words, uppercases and some dots but when i try to save edit github gives:

"Looks like something went wrong!
We track these errors automatically, but if the problem persists feel free to contact us. In the meantime, try refreshing."

TheDevMinerTV:
[GitHub is currently having issues ^^](https://www.githubstatus.com/)

rammlebubble:
successfully sent [2964218], waiting git "Some checks haven’t completed yet"

rammlebubble:
> Why do you think it doesn't work? In the current code, ESP doesn't directly communicate with MPU9250's magnetometer, so there shouldn't be any issues.

Because we tried that, so much ways with you on discord talked about that, tried your fork other's forks and as you see on that forum link it does not working, also opened that forum thread. While its on DMP and connected to esp01 it just does not works cant gets data.

deiteris:
The code was updated to fix this issue within https://github.com/SlimeVR/SlimeVR-Tracker-ESP/pull/108. It shouldn't be an issue anymore.

rammlebubble:
We did conversation over Discord again, I can not try esp01 and 9250 combination right now but I believe it works. (Re-open if there's still this issue happens)

PR #134, opened 2022-03-23
Rework configuration to use LittleFS

TheDevMinerTV:
Recalibrate your IMUs! :D

### DIFF
--- a/ci/build.py
+++ b/ci/build.py
@@ -2,6 +2,7 @@ import json
 import os
 import shutil
 from enum import Enum
+from textwrap import dedent
 from typing import List
 
 COLOR_ESC = '\033['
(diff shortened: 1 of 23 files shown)


kitlith:
will this migrate existing on-flash configuration to littlefs on first run? (i.e. if flash is not already formatted as littlefs, check for existing configuration, then migrate?)

TheDevMinerTV:
no, but this could be added in the future

Eirenliel:
Don't see the reason, jsut re-calibrate :P

TheDevMinerTV:
As I realized a couple hours ago: not even possible, LittleFS is in the exact same space as the EEPROM, so formatting the LittleFS basically wipes the existing configuration from existance.

kitlith:
> As I realized a couple hours ago: not even possible, LittleFS is in the exact same space as the EEPROM, so formatting for LittleFS wipes the existing configuration from existance. 🤷‍♀️

That's whyvyou check for the pre-existing config *before* formatting.

that said, Eiren isn't concerned, so meh.

TinyOra:
This really solved the problem of my 9250 : a certain probability that it will not work on startup( not sending data)
Thanks guys

PR #55, opened 2021-12-18
Support dual MPU9250, change to dmp based fusion from mahony.

Boronare:
Calibration code from #32 : fixed memory leakage, etc...

Support multi MPUs.

Using dmp like MPU6050 with magnetic calibration based on direction cosine matrix.

Magnetic calibration can be tick-skipped, so it has close to 100 TPS.

Issue : Have to wait for reaching absolute rotation. Need initializing code to be added.
Issue2 : How to invoke calibration? I used "Turn on while flipped back, then flip again" to invoke calibration.

PR #135, opened 2022-03-29
MPU9250: Calibration Routine Optimization

LETS-BEE:
and fix some register.
and... reduce memory.
mpu9250 main, mpu6500 aux. this can finish the calibration process.


src/sensors/mpu9250sensor.cpp
```diff
@@ -1,6 +1,6 @@
 /*
     SlimeVR Code is placed under the MIT license
-    Copyright (c) 2021 Eiren Rain, S.J. Remington
+    Copyright (c) 2022 Eiren Rain, S.J. Remington, SlimeVR contributors
```
TheDevMinerTV, line 3: You can leave the year the same

src/sensors/mpu9250sensor.cpp
```diff
@@ -214,6 +217,44 @@ void MPU9250Sensor::getMPUScaled()
...
+    }
+    m_Logger.debug("Calculating calibration data...");
+
+    float M_BAinv[4][3];
+    CalculateCalibration(calibrationDataMag, calibrationSamples, M_BAinv);
+    free(calibrationDataMag);
+
+    m_Logger.debug("[INFO] Magnetometer calibration matrix:");
+    m_Logger.debug("{");
+    for (int i = 0; i < 3; i++) {
```
TheDevMinerTV, line 248: Please format, either you put `{` on the next line or on the current line, not both.

src/sensors/mpu9250sensor.cpp
```diff
@@ -168,14 +168,12 @@ void MPU9250Sensor::getMPUScaled()
     float temp[3];
     int i;
     int16_t ax,ay,az,gx,gy,gz,mx,my,mz;
```
TheDevMinerTV, line 170: You could move these variables (`ax,ay,az,gx,gy,gz,mx,my,mz`) into the `if defined(_MAHONY_H_) || defined(_MADGWICK_H_)` to save even more stack, I believe.

TheDevMinerTV (changes requested):

src/sensors/mpu9250sensor.cpp
LETS-BEE: I didn't code that. that is original code. but I'll fix it.

src/sensors/mpu9250sensor.cpp
LETS-BEE: oh, I fixed it in my another code. I missed this code. thanks.

TheDevMinerTV:
Please merge this: https://github.com/LETS-BEE/SlimeVR-Tracker-ESP-MPU9250/pull/1

src/sensors/mpu9250sensor.cpp
TheDevMinerTV, line 3: still not fixed

src/sensors/mpu9250sensor.cpp
```diff
@@ -167,15 +166,13 @@ void MPU9250Sensor::getMPUScaled()
 {
     float temp[3];
     int i;
+
+#if defined(_MAHONY_H_) || defined(_MADGWICK_H_)
     int16_t ax,ay,az,gx,gy,gz,mx,my,mz;
```
TheDevMinerTV, line 171:
```suggestion
    int16_t ax, ay, az, gx, gy, gz, mx, my, mz;
```

src/sensors/mpu9250sensor.cpp
TheDevMinerTV, line 3:
```suggestion
    Copyright (c) 2021 Eiren Rain, S.J. Remington, SlimeVR contributors
```

TheDevMinerTV (changes requested):

src/sensors/mpu9250sensor.cpp
LETS-BEE: okay okay. Does the code in the file need to be consistent? I'll fix mpu9250sensors.cpp file

src/sensors/mpu9250sensor.cpp
LETS-BEE: actually MPU9250 is working from 30/1/2022, #55 . but who didn't care that. so, I think 2022 is right. however I don't know why fix 2021. can you give me more information?

Eirenliel: The copyright year is placed either when it's first being made or as a range of when it was made. So it's either 2021 or 2021-2022.

src/sensors/mpu9250sensor.cpp
TheDevMinerTV: If you copy stuff, you should also fix the style inconsistencies in the code that you are copying.

LETS-BEE: that seems like an implicit rule. I didn't know that. thaks to you.
if i edit or copy the code, i will remember your comment.

PR #139, opened 2022-03-31
Fix battery sense values

Vyolex:
Corrected values for battery sense, really should be moved to defines.h since these are likely board dependent.


unlogisch04 (approved):
Approve
You can add this values in the define.h i also would suggest it there. This here is only the default value as a fallback and backwards compatibility

TheDevMinerTV (approved):

PR #140, opened 2022-04-01
Move LED define to global + defines to fix TTGO compile

### LINKED ISSUE #37: Compilation error for TTGO T7 V1.3
image

TTGO T7 V1.3 framework defines LED_BUILTIN const instead of definition, this causes the conflict with LED_BUILTIN definition in consts.h in SlimeVR firmware. We may want to add this board type specifically and create separate set of definitions for it.

unlogisch04:
the LED_PIN is wrong in const.h
moved it to globals.h with the option to accept changes from defines.h
with that it should fix #37 

my understanding is, that globals.h adds all the missing stuff from defines.h with the default values.
Any meaning on this?

### DIFF
--- a/src/defines.h
+++ b/src/defines.h
@@ -47,6 +47,14 @@
 // #define BATTERY_SHIELD_R1 100 // Board voltage divider resistor Ain to GND in kOhm
 // #define BATTERY_SHIELD_R2 220 // Board voltage divider resistor Ain to INPUT_BOARD in kOhm
 
+// LED configuration:
+//   LED_PIN
+//     - Number or Symbol (D1,..) of the Output
+//     - To turn off the LED, set LED_PIN to LED_OFF
+//   LED_INVERTED 
+//     - false for output 3.3V on high
+//     - true for pull down to GND on high
+
 // Board-specific configurations
 #if BOARD == BOARD_SLIMEVR
   #define PIN_IMU_SDA 14
@@ -54,31 +62,40 @@
   #define PIN_IMU_INT 16
   #define PIN_IMU_INT_2 13
   #define PIN_BATTERY_LEVEL 17
+  #define LED_PIN LED_BUILTIN
+  #define LED_INVERTED true
 #elif BOARD == BOARD_SLIMEVR_LEGACY || BOARD == BOARD_SLIMEVR_DEV
   #define PIN_IMU_SDA 4
   #define PIN_IMU_SCL 5
   #define PIN_IMU_INT 10
   #define PIN_IMU_INT_2 13
   #define PIN_BATTERY_LEVEL 17
+  #define LED_PIN LED_BUILTIN
+  #define LED_INVERTED true
 #elif BOARD == BOARD_NODEMCU || BOARD == BOARD_WEMOSD1MINI
   #define PIN_IMU_SDA D2
   #define PIN_IMU_SCL D1
   #define PIN_IMU_INT D5
   #define PIN_IMU_INT_2 D6
   #define PIN_BATTERY_LEVEL A0
+  #define LED_PIN LED_BUILTIN
+  #define LED_INVERTED true
 #elif BOARD == BOARD_ESP01
   #define PIN_IMU_SDA 2
   #define PIN_IMU_SCL 0
   #define PIN_IMU_INT 255
   #define PIN_IMU_INT_2 255
-  #define ENABLE_LEDS false
   #define PIN_BATTERY_LEVEL 255
+  #define LED_PIN LED_OFF
+  #define LED_INVERTED false
 #elif BOARD == BOARD_TTGO_TBASE
   #define PIN_IMU_SDA 5
   #define PIN_IMU_SCL 4
   #define PIN_IMU_INT 14
   #define PIN_IMU_INT_2 13
   #define PIN_BATTERY_LEVEL A0
+  #define LED_PIN LED_BUILTIN
+  #define LED_INVERTED true
 #elif BOARD == BOARD_CUSTOM
   // Define pins by the examples above
 #elif BOARD == BOARD_WROOM32
@@ -87,4 +104,6 @@
   #define PIN_IMU_INT 23
   #define PIN_IMU_INT_2 25
   #define PIN_BATTERY_LEVEL 36
+  #define LED_PIN LED_BUILTIN
+  #define LED_INVERTED false
 #endif
--- a/src/globals.h
+++ b/src/globals.h
@@ -40,7 +40,32 @@
 #define BATTERY_MONITOR BAT_INTERNAL
 #endif
 
-#if defined(LED_INVERTED) && LED_INVERTED
+// If LED_PIN is not defined in "defines.h" take the default pin from "pins_arduino.h" framework.
+// If there is no pin defined for the board, use LED_PIN 255 and disable LED
+#if defined(LED_PIN)
+    // LED_PIN is defined
+    #if (LED_PIN < 0) || (LED_PIN >= LED_OFF)
+        #define ENABLE_LEDS false    
+    #else
+        #define ENABLE_LEDS true
+    #endif
+#else
+    // LED_PIN is not defined
+    #if defined(LED_BUILTIN)
+        #define LED_PIN LED_BUILTIN
+        #define ENABLE_LEDS true
+    #else
+        #define LED_PIN LED_OFF
+        #define ENABLE_LEDS false
+    #endif
+#endif
+
+#if !defined(LED_INVERTED)
+    // default is inverted for SlimeVR / ESP-12E
+    #define LED_INVERTED true
+#endif
+
+#if LED_INVERTED
 #define LED_ON LOW
 #define LED_OFF HIGH
 #else
--- a/src/consts.h
+++ b/src/consts.h
@@ -49,6 +49,8 @@
 #define BAT_MCP3021 3
 #define BAT_INTERNAL_MCP3021 4
 
+#define LED_OFF 255
+
 #define POWER_SAVING_LEGACY 0 // No sleeping, but PS enabled
 #define POWER_SAVING_NONE 1 // No sleeping, no PS => for connection issues
 #define POWER_SAVING_MINIMUM 2 // Sleeping and PS => default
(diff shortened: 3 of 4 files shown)


src/globals.h
```diff
@@ -40,7 +40,27 @@
 #define BATTERY_MONITOR BAT_INTERNAL
 #endif
 
-#if defined(LED_INVERTED) && LED_INVERTED
+// If LED_PIN is not defined in "defines.h" take the default pin from "pins_arduino.h" framework.
+// If there is no pin defined for the board, use LED_PIN 255 and disable LED
+#if !defined(LED_PIN)
+#if defined(LED_BUILTIN)
```
TheDevMinerTV, line 46:
```suggestion
    #if defined(LED_BUILTIN)
```

src/globals.h
```diff
@@ -40,7 +40,27 @@
...
+#define LED_PIN LED_BUILTIN
```
TheDevMinerTV, line 47:
```suggestion
        #define LED_PIN LED_BUILTIN
```

Eirenliel:
It's not wrong for ESP-12f: https://randomnerdtutorials.com/esp8266-pinout-reference-gpios/

src/globals.h
```diff
@@ -40,7 +40,27 @@
...
+#define ENABLE_LEDS true
```
TheDevMinerTV, line 48:
```suggestion
        #define ENABLE_LEDS true
```

Also, what if `ENABLE_LEDS` is set to `false` in `defines.h`?

src/globals.h
```diff
@@ -40,7 +40,27 @@
...
+#else
+#define LED_PIN 255
+#define ENABLE_LEDS false
+#endif
+#else
+// LED_PIN is defined
+#define ENABLE_LEDS true
```
TheDevMinerTV, line 55:
```suggestion
    #define ENABLE_LEDS true
```

src/globals.h
```diff
@@ -40,7 +40,27 @@
...
+#if !defined(LED_PIN)
+#if defined(LED_BUILTIN)
+#define LED_PIN LED_BUILTIN
+#define ENABLE_LEDS true
+#else
+#define LED_PIN 255
+#define ENABLE_LEDS false
+#endif
+#else
+// LED_PIN is defined
```
TheDevMinerTV, line 54:
```suggestion
    // LED_PIN is defined
```

src/globals.h
```diff
@@ -40,7 +40,27 @@
...
 #endif
 
-#if defined(LED_INVERTED) && LED_INVERTED
+// If LED_PIN is not defined in "defines.h" take the default pin from "pins_arduino.h" framework.
+// If there is no pin defined for the board, use LED_PIN 255 and disable LED
+#if !defined(LED_PIN)
+#if defined(LED_BUILTIN)
+#define LED_PIN LED_BUILTIN
+#define ENABLE_LEDS true
+#else
```
TheDevMinerTV, line 49:
```suggestion
    #else
```

src/globals.h
```diff
@@ -40,7 +40,27 @@
...
+#define LED_PIN 255
```
TheDevMinerTV, line 50:
```suggestion
        #define LED_PIN 255
```

src/globals.h
```diff
@@ -40,7 +40,27 @@
...
+#define ENABLE_LEDS false
+#endif
```
TheDevMinerTV, line 52:
```suggestion
    #endif
```

src/globals.h
```diff
@@ -40,7 +40,27 @@
...
-#if defined(LED_INVERTED) && LED_INVERTED
+// If LED_PIN is not defined in "defines.h" take the default pin from "pins_arduino.h" framework.
+// If there is no pin defined for the board, use LED_PIN 255 and disable LED
+#if !defined(LED_PIN)
+#if defined(LED_BUILTIN)
+#define LED_PIN LED_BUILTIN
+#define ENABLE_LEDS true
+#else
+#define LED_PIN 255
+#define ENABLE_LEDS false
```
TheDevMinerTV, line 51:
```suggestion
        #define ENABLE_LEDS false
```

src/globals.h
```diff
@@ -40,7 +40,27 @@
...
+#endif
+#else
+// LED_PIN is defined
+#define ENABLE_LEDS true
+#endif
+
+#if !defined(LED_INVERTED)
+// default is inverted for SlimeVR / ESP-12F
+#define LED_INVERTED true
```
TheDevMinerTV, line 60:
```suggestion
    #define LED_INVERTED true
```

src/globals.h
```diff
@@ -40,7 +40,27 @@
...
+#define LED_PIN 255
+#define ENABLE_LEDS false
+#endif
+#else
+// LED_PIN is defined
+#define ENABLE_LEDS true
+#endif
+
+#if !defined(LED_INVERTED)
+// default is inverted for SlimeVR / ESP-12F
```
TheDevMinerTV, line 59:
```suggestion
    // default is inverted for SlimeVR / ESP-12F
```

TheDevMinerTV (changes requested):

Eirenliel:
These suggestings aren't helpful... Why do we need all of this? :nya_a: They probably should be part of the board definition idk. LED_BUILTIN should be defined for ESP from the toolset if i'm right anyway?

TheDevMinerTV:
The TTGO things implement `LED_BUILTINT` as `static constexpr LED_BUILTIN` which collides with on of our defines so that the code basically doesn't compile for the TTGO boards, the ESPs on the other hand just use a `#define` for this.

Eirenliel:
I see, I hate it. That's still should be part of board definition, no? Also I think that's not TTGO, that's ESP32 behaviour?

unlogisch04:
I want the most flexibility for the defines.h
so everyone can override the default pins in defines.h or not if he wants to stay with the board (Arduino framework definition)
I did try to stay compatible with the generator, so it works as it would work with the old defines.h

I mean not the number or value of it. For esp-12f that value is already defined in the Arduino framework.
It is just in the wrong spot in the config .

1. const.h gets included as fist.
2. (arduino_board_default)
3. defines.h 
4. globals.h
globals.h is referenced as endpoint for the values. (the defines generated from here will be used in the code)

unlogisch04:
Updated: Now it can be part of the defines.h #define board config

src/globals.h
```diff
@@ -40,7 +40,27 @@
 #define BATTERY_MONITOR BAT_INTERNAL
 #endif
 
-#if defined(LED_INVERTED) && LED_INVERTED
+// If LED_PIN is not defined in "defines.h" take the default pin from "pins_arduino.h" framework.
+// If there is no pin defined for the board, use LED_PIN 255 and disable LED
+#if !defined(LED_PIN)
+    #if defined(LED_BUILTIN)
+        #define LED_PIN LED_BUILTIN
+        #define ENABLE_LEDS true
```
TheDevMinerTV, line 48: What if `ENABLE_LEDS` is set to `false`?

unlogisch04: Good catch. I think i have to redo that whole part, it it should not get more complex.

unlogisch04: I changed some stuff.  Can you review again?
I added for every board. Ignoring ENABLE_LEDS  (overwriting it)
  #define LED_PIN LED_BUILTIN
  #define LED_INVERTED true

TheDevMinerTV (changes requested):

src/defines.h
```diff
@@ -47,38 +47,52 @@
 // #define BATTERY_SHIELD_R1 100 // Board voltage divider resistor Ain to GND in kOhm
 // #define BATTERY_SHIELD_R2 220 // Board voltage divider resistor Ain to INPUT_BOARD in kOhm
 
+// LED configuration; uncomment if you need to change it
```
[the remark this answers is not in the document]
unlogisch04: Valid. Would that be better 
```
// LED configuration:
//   LED_PIN
//     - Number or Symbol (D1,..) of the Output
//     - To turn off the LED, set LED_PIN to LED_OFF
//   LED_INVERTED 
//     - false for output 3.3V on high
//     - true for pull down to GND on high
```

src/globals.h
```diff
@@ -40,7 +40,31 @@
...
+    #else
+        #define LED_PIN 255
```
deiteris, line 50:
```suggestion
        #define LED_PIN LED_OFF
```

src/defines.h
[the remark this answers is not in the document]
deiteris: Yeah, the instructions would be better in this case.

src/globals.h
```diff
@@ -40,7 +40,31 @@
...
+        #define ENABLE_LEDS false
+    #endif
+#else
+    // LED_PIN is defined
+    #if (LED_PIN < 0) || (LED_PIN >= 255)
```
TheDevMinerTV, line 55:
```suggestion
    #if (LED_PIN < 0) || (LED_PIN >= LED_OFF)
```

TheDevMinerTV (changes requested):

src/globals.h
```diff
@@ -40,7 +40,31 @@
...
+    // LED_PIN is defined
+    #if (LED_PIN < 0) || (LED_PIN >= LED_OFF)
+        #define ENABLE_LEDS false    
+    #else
+        #define ENABLE_LEDS true
+    #endif
+#endif
+
+#if !defined(LED_INVERTED)
+    // default is inverted for SlimeVR / ESP-12F
```
TheDevMinerTV, line 63:
```suggestion
    // default is inverted for SlimeVR / ESP-12E
```

src/globals.h
```diff
@@ -40,7 +40,31 @@
 #define BATTERY_MONITOR BAT_INTERNAL
 #endif
 
-#if defined(LED_INVERTED) && LED_INVERTED
+// If LED_PIN is not defined in "defines.h" take the default pin from "pins_arduino.h" framework.
+// If there is no pin defined for the board, use LED_PIN 255 and disable LED
+#if !defined(LED_PIN)
```
TheDevMinerTV, line 45: We should probably invert these branches, just to make it look a bit neater.

TheDevMinerTV (changes requested):

src/globals.h
unlogisch04: ok done

TheDevMinerTV (approved):

Eirenliel:
wow what happened here... good job y'all :D

PR #141, opened 2022-04-01
Some improvements to ICM support.

ThreadOfFate:
6Dof mode no longer tries to save compass bias.
Improve motion loop, by now only set most recent data.


TheDevMinerTV:
Time to rebase #134 ...

PR #143, opened 2022-04-02
Increase MPU9250 magnetometer sample rate to configured TPS

TheDevMinerTV:


rickwinter2:
Was talking to the Swiss guy on his new video. Esp32 da maybe a better connection  btw probably commenting in the wrong bit here but hey ho 😀

PR #145, opened 2022-04-03
Added option to maximize update rate of ICM on supported boards.

ThreadOfFate:
Tested on esp8266.
Left off by default as battery life will likely be more noticeable than performance increases.


src/defines.h
```diff
@@ -47,6 +48,13 @@
 // #define BATTERY_SHIELD_R1 100 // Board voltage divider resistor Ain to GND in kOhm
 // #define BATTERY_SHIELD_R2 220 // Board voltage divider resistor Ain to INPUT_BOARD in kOhm
 
+//Only works in esp8266
+#if BOARD == BOARD_WEMOSD1MINI
+#define MAX_ODR false // (True) enables Fastest Output Rate (False) caps it to 100Hz (ICM20948 only)
```
Eirenliel, line 53: Why is it supported only on WEMOSD1MINI and not on all ESP8266 or actually all boards? Also, defines.h is for defines not for processing, checks should be done in globals.h

src/debug.h
```diff
@@ -33,6 +33,7 @@
 #define BIAS_DEBUG false // Printing BIAS Variables to serial (ICM20948 only)
 #define ENABLE_TAP false // monitor accel for (triple) tap events and send them. Uses more cpu, disable if problems. Server does nothing with value so disabled atm
 
+
```
Eirenliel, line 36: ?

src/sensors/icm20948sensor.cpp
```diff
@@ -210,7 +210,32 @@ void ICM20948Sensor::motionSetup() {
     }
 
     // Might need to set up other DMP functions later, just Quad6/Quad9 for now
-
+    #if MAX_ODR
+    if (USE_6_AXIS)
```
TheDevMinerTV, line 214: While you're here, might as well also rewrite those `if`s into preprocessor `#if`s

src/sensors/icm20948sensor.cpp
```diff
@@ -210,7 +210,32 @@ void ICM20948Sensor::motionSetup() {
...
+            m_Logger.debug("Set Quat9 to Max frequency");
+        }
+        else
+        {
+           m_Logger.fatal("Failed to set Quat9 to Max frequency");
+            return;
+        }
+    }
+    #else
     if (USE_6_AXIS)
```
TheDevMinerTV, line 239: Would be sweet if you'd also rewrite this one into a `#if` :D

TheDevMinerTV (changes requested):
Same things that Eiren said + some extra

TheDevMinerTV:
@ThreadOfFate please rebase against `main` and look at the reviews.

PR #153, opened 2022-04-22
Fix i2cscan trying to begin in slave mode

sctanf:
~Not working, causes esp to crash(?) mid-scan~
Works fine with only i2c wires attached

### DIFF
--- a/lib/i2cscan/i2cscan.cpp
+++ b/lib/i2cscan/i2cscan.cpp
@@ -1,10 +1,10 @@
 #include "i2cscan.h"
 
 #ifdef ESP8266
-uint8_t portArray[] = {16, 5, 4, 2, 14, 12, 13};
+int portArray[] = {16, 5, 4, 2, 14, 12, 13};
 String portMap[] = {"D0", "D1", "D2", "D4", "D5", "D6", "D7"};
 #elif defined(ESP32)
-uint8_t portArray[] = {4, 13, 14, 15, 16, 17, 18, 19, 21, 22, 23, 25, 26, 27, 32, 33};
+int portArray[] = {4, 13, 14, 15, 16, 17, 18, 19, 21, 22, 23, 25, 26, 27, 32, 33};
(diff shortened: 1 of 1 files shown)


lib/i2cscan/i2cscan.cpp
```diff
@@ -1,10 +1,10 @@
 #include "i2cscan.h"
 
 #ifdef ESP8266
-uint8_t portArray[] = {16, 5, 4, 2, 14, 12, 13};
+int portArray[] = {16, 5, 4, 2, 14, 12, 13};
```
TheDevMinerTV, line 4: explain your reason to change this, there's no pins < 0

lib/i2cscan/i2cscan.cpp
```diff
@@ -1,10 +1,10 @@
...
 String portMap[] = {"D0", "D1", "D2", "D4", "D5", "D6", "D7"};
 #elif defined(ESP32)
-uint8_t portArray[] = {4, 13, 14, 15, 16, 17, 18, 19, 21, 22, 23, 25, 26, 27, 32, 33};
+int portArray[] = {4, 13, 14, 15, 16, 17, 18, 19, 21, 22, 23, 25, 26, 27, 32, 33};
```
TheDevMinerTV, line 7: same here.

TheDevMinerTV (changes requested):
found these issues things

TheDevMinerTV (approved):

lib/i2cscan/i2cscan.cpp
```diff
@@ -81,6 +81,7 @@ namespace I2CSCAN
                 Serial.println(address, HEX);
             }
         }
+        Wire.end();
```
unlogisch04, line 84: Wire.end(); does not exist in ESP8266
`#ifdef ESP32`
`Wire.end()`
 `#endif`

sctanf:
Not an issue since trackers are using espressif32 v3.5.0 for now

PR #154, opened 2022-04-23
Fixing compile Error if LED_BUILTIN is not defined

unlogisch04:
Instead of using LED_BUILTIN as a value in defines.h
we can leave LED_PIN undefined. In that case global.h 
will look if LED_BUILTIN is defined or not.


TheDevMinerTV (approved):
LGTM

PR #156, opened 2022-04-24
Add required variable for flashing with no changes

IlexisNakamori:


Eirenliel:
This is not required. It should find the port automatically, and it can be different for everyone.

PR #159, opened 2022-04-27
Unify MPU libraries

deiteris:
Deduplicate and remove unused code. Since register maps for mpu6x00 and mpu60x0 are mostly identical and DMP is applicable to the entire family, the use of separate libraries doesn't make any sense. This also removes unused data from DMP packet that should improve reading speed.

Needs testing.


inkuxuan:
If merged please add also 0x74 as a possible device ID for MPU6500. It is found that some of the MPU6500s uses this address.

unlogisch04:
Works on BMI160, MPU6050c, ICM20948. I did not test it as tracker itself

nekomona:
For the DMP packet content part, now the server by default uses acceleration data to determine taps and foot compensations, so it might be a better idea to remove the gyro output but keep the acceleration ones

PR #162, opened 2022-06-04
feat: multiple wifi credentials

ZigTag:
I keep having to reset the credentials because I have multiple play spaces with two different login credentials, so I decided to make this to make my problem more bearable


TheDevMinerTV:
I already have a better solution on hand, give me a few days to polish it up though :)

ZigTag:
Would it be wise to reconsider this patch? Seems like #164 wasn't viable

TheDevMinerTV:
Only if you're up to rewriting the WifiHandler into a class and refactoring some of the code in there (you can copy paste most stuff from my PR, probably).

ZigTag:
Yeah I'd make it work on the newest versions, just asking if its worth the effort

ZigTag:
Just to note, this patch is multiple HARDCODED wifi credentials, not multiple dynamic ones

PR #163, opened 2022-06-05
Fix esp32c3 build

TheButlah:


src/configuration/Configuration.cpp
```diff
@@ -21,15 +21,7 @@
...
-#ifdef ESP32
-    // #define CONFIG_LITTLEFS_SPIFFS_COMPAT 1
-    // #define CONFIG_LITTLEFS_LFS_YES_TRACE
-    // #define LFS_YES_TRACE
-    #include <LITTLEFS.h>
-    #define LittleFS LITTLEFS
-#else
-    #include <LittleFS.h>
-#endif
+#include <LittleFS.h>
```
TheDevMinerTV, line 24: This isn't gonna work with the normal ESP32 builds

TheDevMinerTV (changes requested):

src/configuration/Configuration.cpp
TheButlah: how come?

unlogisch04:
fixed with #170

TheButlah:
> fixed with #170

Are we sure? I thought Anna said that it didn't compile for her or something

src/configuration/Configuration.cpp
TheDevMinerTV: because you're not bumping the framework version for the "old" ESP32, in which this file doesn't exist. Which is why we had to use a library to provide this, which needs some defines to work properly.

TheButlah: Is it resolved with the latest commits?

TheButlah:
> We should probably add a CI step for the C3 as well (just edit the JSON file in `ci/`)

Can you make this change?

TheDevMinerTV:
Welp, as long as this doesn't compile you can forget about merging

Eirenliel:
Wow Anna CI-blocked Butlah :nya_a:

TheButlah:
Its the `-fpermissive` bug, I thought you said that was fixed @unlogisch04

TheButlah:
To be fair, the `-fpermissive` has been a problem for a while and its only showing up now due to the updated compiler

TheDevMinerTV:
> Wow Anna CI-blocked Butlah :nya_a:

the CI promises to fix esp32c3 builds but doesn't even build it in CI smh
if it doesn't compile in this PR it shouldn't get merged. as easy as that.

TheDevMinerTV:
@TheButlah please rebase against `main` :)

TheDevMinerTV:
I think that changing the types of the configuration file would've been smarter.
That way we:
- Change the library, therefore making updating the lib easier
- Have less changes in general

unlogisch04:
I did not find this bias function in a actual upstream lib. So not sure if this was added later anyway. I look at all sparkfun icm20948 forks, that where ahead of the original.

TheDevMinerTV (approved):

unlogisch04 (changes requested):
Rebase, and we need to change https://github.com/SlimeVR/SlimeVR-Tracker-ESP/blob/main/platformio-tools.ini now too.

unlogisch04 (approved):

Eirenliel (approved):

PR #166, opened 2022-06-11
Refactor BNO08x conditions

deiteris:
Replaces runtime conditions with compile time conditions

### DIFF
--- a/src/debug.h
+++ b/src/debug.h
@@ -26,8 +26,9 @@
 #include "logging/Level.h"
 
 #define IMU_MPU6050_RUNTIME_CALIBRATION // Comment to revert to startup/traditional-calibration
-#define BNO_USE_ARVR_STABILIZATION // Comment to not use stabilization for BNO085+ IMUs
-#define USE_6_AXIS true // uses 9 (with mag) if false (only for ICM-20948 currently)
+#define BNO_USE_ARVR_STABILIZATION true // Set to false to disable stabilization for BNO085+ IMUs
+#define BNO_USE_MAGNETOMETER_CORRECTION false // Set to true to enable magnetometer correction for BNO08x IMUs. Only works with USE_6_AXIS set to true.
+#define USE_6_AXIS true // uses 9 (with mag) if false (only for ICM-20948 and BNO08x currently)
 #define LOAD_BIAS 1 // Loads the bias values from NVS on start (ESP32 Only)
 #define SAVE_BIAS 1 // Periodically saves bias calibration data to NVS (ESP32 Only)
 #define BIAS_DEBUG false // Printing BIAS Variables to serial (ICM20948 only)
--- a/src/sensors/bno080sensor.cpp
+++ b/src/sensors/bno080sensor.cpp
@@ -54,12 +54,12 @@ void BNO080Sensor::motionSetup()
 
 #if USE_6_AXIS == true
 #if (IMU == IMU_BNO085 || IMU == IMU_BNO086) && BNO_USE_ARVR_STABILIZATION
-            imu.enableARVRStabilizedGameRotationVector(10);
+    imu.enableARVRStabilizedGameRotationVector(10);
 #else
-            imu.enableGameRotationVector(10);
+    imu.enableGameRotationVector(10);
 #endif
 #if BNO_USE_MAGNETOMETER_CORRECTION == true
-                imu.enableRotationVector(1000);
+    imu.enableRotationVector(1000);
 #endif
 #else
 #if (IMU == IMU_BNO085 || IMU == IMU_BNO086) && BNO_USE_ARVR_STABILIZATION
@@ -115,16 +115,16 @@ void BNO080Sensor::motionLoop()
         if (imu.hasNewGameQuat())
         {
             imu.getGameQuat(quaternion.x, quaternion.y, quaternion.z, quaternion.w, calibrationAccuracy);
-                quaternion *= sensorOffset;
+            quaternion *= sensorOffset;
 
 #if ENABLE_INSPECTION
-                {
-                    Network::sendInspectionFusedIMUData(sensorId, quaternion);
-                }
+            {
+                Network::sendInspectionFusedIMUData(sensorId, quaternion);
+            }
 #endif // ENABLE_INSPECTION
 
-                    newData = true;
-                }
+            newData = true;
+        }
 #if BNO_USE_MAGNETOMETER_CORRECTION == true
         if (imu.hasNewMagQuat())
         {
@@ -132,24 +132,24 @@ void BNO080Sensor::motionLoop()
             magQuaternion *= sensorOffset;
 
 #if ENABLE_INSPECTION
-                {
+            {
                 Network::sendInspectionCorrectionData(sensorId, quaternion);
-                }
+            }
 #endif // ENABLE_INSPECTION
 
             newMagData = true;
-            }
+        }
 #endif // BNO_USE_MAGNETOMETER_CORRECTION
 #else // USE_6_AXIS
         if (imu.hasNewQuat())
-            {
+        {
             imu.getQuat(quaternion.x, quaternion.y, quaternion.z, quaternion.w, magneticAccuracyEstimate, calibrationAccuracy);
             quaternion *= sensorOffset;
 
 #if ENABLE_INSPECTION
-                {
+            {
                 Network::sendInspectionFusedIMUData(sensorId, quaternion);
-                }
+            }
 #endif // ENABLE_INSPECTION
 
             if (!OPTIMIZE_UPDATES || !lastQuatSent.equalsWithEpsilon(quaternion))
@@ -171,8 +171,6 @@ void BNO080Sensor::motionLoop()
             imu.getAccel(v[0], v[1], v[2], acc);
             Network::sendAccel(v, PACKET_ACCEL);
         }
-        if (m_IntPin == 255 || imu.I2CTimedOut())
-            break;
     }
     if (lastData + 1000 < millis() && configured)
     {
@@ -210,7 +208,7 @@ void BNO080Sensor::sendData()
         newData = false;
         Network::sendRotationData(&quaternion, DATA_TYPE_NORMAL, calibrationAccuracy, sensorId);
 #if USE_6_AXIS == false
-            Network::sendMagnetometerAccuracy(magneticAccuracyEstimate, sensorId);
+        Network::sendMagnetometerAccuracy(magneticAccuracyEstimate, sensorId);
 #endif
 
 #ifdef DEBUG_SENSOR
--- a/src/sensors/bno080sensor.cpp
+++ b/src/sensors/bno080sensor.cpp
@@ -52,27 +52,23 @@ void BNO080Sensor::motionSetup()
                   imu.swVersionPatch
                 );
 
-    bool useStabilization = false;
-#ifdef BNO_USE_ARVR_STABILIZATION
-    useStabilization = true;
-#endif
-    if(useStabilization && (sensorType == IMU_BNO085 || sensorType == IMU_BNO086)) {
-        if(useMagnetometerAllTheTime) {
-            imu.enableARVRStabilizedRotationVector(10);
-        } else {
+#if USE_6_AXIS == true
+#if (IMU == IMU_BNO085 || IMU == IMU_BNO086) && BNO_USE_ARVR_STABILIZATION
             imu.enableARVRStabilizedGameRotationVector(10);
-            if (useMagnetometerCorrection)
-                imu.enableRotationVector(1000);
-        }
-    } else {
-        if(useMagnetometerAllTheTime) {
-            imu.enableRotationVector(10);
-        } else {
+#else
             imu.enableGameRotationVector(10);
-            if(useMagnetometerCorrection)
+#endif
+#if BNO_USE_MAGNETOMETER_CORRECTION == true
                 imu.enableRotationVector(1000);
-        }
-    }
+#endif
+#else
+#if (IMU == IMU_BNO085 || IMU == IMU_BNO086) && BNO_USE_ARVR_STABILIZATION
+    imu.enableARVRStabilizedRotationVector(10);
+#else
+    imu.enableRotationVector(10);
+#endif
+#endif
+
     imu.enableTapDetector(100);
 
 #if ENABLE_INSPECTION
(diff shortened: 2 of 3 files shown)


Eirenliel (approved):

src/sensors/bno080sensor.cpp
```diff
@@ -52,27 +52,23 @@ void BNO080Sensor::motionSetup()
                   imu.swVersionPatch
                 );
 
-    bool useStabilization = false;
-#ifdef BNO_USE_ARVR_STABILIZATION
-    useStabilization = true;
+#if USE_6_AXIS == true
```
TheDevMinerTV, line 55:
```suggestion
#if USE_6_AXIS
```

src/sensors/bno080sensor.cpp
```diff
@@ -115,55 +111,55 @@ void BNO080Sensor::motionLoop()
 
         lastReset = 0;
         lastData = millis();
-        if (useMagnetometerAllTheTime || !useMagnetometerCorrection)
+#if USE_6_AXIS == true
```
TheDevMinerTV, line 114:
```suggestion
#if USE_6_AXIS
```

src/sensors/bno080sensor.cpp
```diff
@@ -213,19 +207,22 @@ void BNO080Sensor::sendData()
...
-            Network::sendMagnetometerAccuracy(magneticAccuracyEstimate, sensorId);
+#if USE_6_AXIS == false
+        Network::sendMagnetometerAccuracy(magneticAccuracyEstimate, sensorId);
+#endif
 
 #ifdef DEBUG_SENSOR
         m_Logger.trace("Quaternion: %f, %f, %f, %f", UNPACK_QUATERNION(quaternion));
 #endif
     }
+#if USE_6_AXIS == true && BNO_USE_MAGNETOMETER_CORRECTION == true
```
TheDevMinerTV, line 218:
```suggestion
#if USE_6_AXIS && BNO_USE_MAGNETOMETER_CORRECTION
```

src/sensors/bno080sensor.cpp
```diff
@@ -213,19 +207,22 @@ void BNO080Sensor::sendData()
     {
         newData = false;
         Network::sendRotationData(&quaternion, DATA_TYPE_NORMAL, calibrationAccuracy, sensorId);
-        if (useMagnetometerAllTheTime)
-            Network::sendMagnetometerAccuracy(magneticAccuracyEstimate, sensorId);
+#if USE_6_AXIS == false
```
TheDevMinerTV, line 210:
```suggestion
#if !USE_6_AXIS
```

src/sensors/bno080sensor.cpp
```diff
@@ -115,55 +111,55 @@ void BNO080Sensor::motionLoop()
...
-                }
+            {
+                Network::sendInspectionFusedIMUData(sensorId, quaternion);
             }
+#endif // ENABLE_INSPECTION
+
+            newData = true;
         }
-        else
+#if BNO_USE_MAGNETOMETER_CORRECTION == true
```
TheDevMinerTV, line 128:
```suggestion
#if BNO_USE_MAGNETOMETER_CORRECTION
```

TheDevMinerTV (changes requested):

src/sensors/bno080sensor.cpp
```diff
@@ -175,8 +171,6 @@ void BNO080Sensor::motionLoop()
             imu.getAccel(v[0], v[1], v[2], acc);
             Network::sendAccel(v, PACKET_ACCEL);
         }
-        if (m_IntPin == 255 || imu.I2CTimedOut())
```
deiteris, line 178: My only concern is this change. It should be redundant because `BNO080::getReadings()` already checks whether INT pin is 255 and uses `waitForI2C()` to when receiving packet to check if I2C is timet out. But pls test.

TheDevMinerTV: @Eirenliel please check if this works. (I don't have BNOs)

Eirenliel: I don't have anything, I live in an airbnb with a laptop 😢

Philip6294: So FWIW, I tried this version with a BNO085 Tracker and it ran fine without any errors at like ~100 TPS when moving the tracker

Eirenliel: This check is specifically so it DOESN'T wait for I2C if there are no INT pin defined, otherwise it would go back to reading until it timed out, and can't do other stuff, like sending data or checking another tracker. So this is needed when int pin is not connected/not defined.

Philip6294: ah okay, 
I just tried it with wrong INT Pins in defines.h (just set both of them to 0).
It resulted in constantly printing Sensor 0 does not respond error. 
An Aux MPU9250 was detected and working tho.

Eirenliel: INT set to 0 is still an INT, try with 255 - disabled.

Eirenliel: Or just return these two lines :D

Philip6294: When INT is set to 255 it runs like normal. The BNO is detected and data is being read.

Eirenliel: It should get low TPS with aux if i understand correctly

Philip6294: Jup, TPS is 100 but sometimes dropping to 60-70 for the BNO and the MPU

deiteris: I suspect that blocking manner of this loop could also be causing these TPS drops... @Eirenliel was it tested before with `if (imu.dataAvailable())` instead of `while (imu.dataAvailable())` by any chance?

Eirenliel: >  The first check it does, is whether INT pin is set to 255, and if not or pin level is LOW it receives the packet until I2C timeout.

That's the point, we don't want this to happen SECOND TIME. We just called dataAvailable(), we don't want to wait for i2c again. That's why it breaks if the INT pin is not used.

Just revert it please and it can be merged.

Eirenliel (approved):

Eirenliel:
@TheDevMinerTV please check if your requests has been resolved and if they did, approve the PR 😋

src/sensors/bno080sensor.cpp
```diff
@@ -115,55 +111,59 @@ void BNO080Sensor::motionLoop()
...
-                    lastQuatSent = quaternion;
-                }
+            if (!OPTIMIZE_UPDATES || !lastQuatSent.equalsWithEpsilon(quaternion))
+            {
+                newData = true;
+                lastQuatSent = quaternion;
             }
         }
-        else
+#if BNO_USE_MAGNETOMETER_CORRECTION
```
TheDevMinerTV, line 132:
```suggestion

#if BNO_USE_MAGNETOMETER_CORRECTION
```

src/sensors/bno080sensor.cpp
```diff
@@ -115,55 +111,59 @@ void BNO080Sensor::motionLoop()
 
         lastReset = 0;
         lastData = millis();
-        if (useMagnetometerAllTheTime || !useMagnetometerCorrection)
+#if USE_6_AXIS
```
TheDevMinerTV, line 114:
```suggestion

#if USE_6_AXIS
```

src/sensors/bno080sensor.cpp
```diff
@@ -213,19 +213,22 @@ void BNO080Sensor::sendData()
...
-            Network::sendMagnetometerAccuracy(magneticAccuracyEstimate, sensorId);
+#if !USE_6_AXIS
+        Network::sendMagnetometerAccuracy(magneticAccuracyEstimate, sensorId);
+#endif
 
 #ifdef DEBUG_SENSOR
         m_Logger.trace("Quaternion: %f, %f, %f, %f", UNPACK_QUATERNION(quaternion));
 #endif
     }
+#if USE_6_AXIS && BNO_USE_MAGNETOMETER_CORRECTION
```
TheDevMinerTV, line 224:
```suggestion

#if USE_6_AXIS && BNO_USE_MAGNETOMETER_CORRECTION
```

src/sensors/bno080sensor.cpp
```diff
@@ -213,19 +213,22 @@ void BNO080Sensor::sendData()
...
     if (newMagData)
     {
         newMagData = false;
         Network::sendRotationData(&magQuaternion, DATA_TYPE_CORRECTION, magCalibrationAccuracy, sensorId);
         Network::sendMagnetometerAccuracy(magneticAccuracyEstimate, sensorId);
     }
+#endif
```
TheDevMinerTV, line 231:
```suggestion
#endif

```

TheDevMinerTV (changes requested):

TheDevMinerTV:
@deiteris as @TheButlah once mentioned to me, please don't resolve the conversations, the person that created the conversation should resolve it. just gets hard to manage :)

src/debug.h
```diff
@@ -26,8 +26,9 @@
 #include "logging/Level.h"
 
 #define IMU_MPU6050_RUNTIME_CALIBRATION // Comment to revert to startup/traditional-calibration
-#define BNO_USE_ARVR_STABILIZATION // Comment to not use stabilization for BNO085+ IMUs
-#define USE_6_AXIS true // uses 9 (with mag) if false (only for ICM-20948 currently)
+#define BNO_USE_ARVR_STABILIZATION true // Set to false to disable stabilization for BNO085+ IMUs
+#define BNO_USE_MAGNETOMETER_CORRECTION false // Set to true to enable magnetometer correction for BNO08x IMUs. Only works with USE_6_AXIS set to true.
+#define USE_6_AXIS true // uses 9 (with mag) if false (only for ICM-20948 and BNO08x currently)
```
TheDevMinerTV, line 31:
```suggestion
#define USE_6_AXIS true // uses 9 DoF (with mag) if false (only for ICM-20948 and BNO08x currently)
```

src/sensors/bno080sensor.cpp
```diff
@@ -52,27 +52,23 @@ void BNO080Sensor::motionSetup()
                   imu.swVersionPatch
                 );
 
-    bool useStabilization = false;
-#ifdef BNO_USE_ARVR_STABILIZATION
-    useStabilization = true;
+#if USE_6_AXIS
+#if (IMU == IMU_BNO085 || IMU == IMU_BNO086) && BNO_USE_ARVR_STABILIZATION
```
TheDevMinerTV, line 56: Maybe we should indent these?

TheDevMinerTV (changes requested):

deiteris:
> @deiteris as @TheButlah once mentioned to me, please don't resolve the conversations, the person that created the conversation should resolve it. just gets hard to manage :)

I don't resolve them myself, they're resolved automatically when I apply suggestions. If there's an option, I could disable it.

src/sensors/bno080sensor.cpp
deiteris: If we should, then how? Like this?
```c
#if USE_6_AXIS
    #if (IMU == IMU_BNO085 || IMU == IMU_BNO086) && BNO_USE_ARVR_STABILIZATION
    imu.enableARVRStabilizedGameRotationVector(10);
    #else
    imu.enableGameRotationVector(10);
    #endif
    #if BNO_USE_MAGNETOMETER_CORRECTION
    imu.enableRotationVector(1000);
    #endif
#else
    #if (IMU == IMU_BNO085 || IMU == IMU_BNO086) && BNO_USE_ARVR_STABILIZATION
    imu.enableARVRStabilizedRotationVector(10);
    #else
    imu.enableRotationVector(10);
    #endif
#endif
```

TheDevMinerTV:
> > @deiteris as @TheButlah once mentioned to me, please don't resolve the conversations, the person that created the conversation should resolve it. just gets hard to manage :)
> 
> I don't resolve them myself, they're resolved automatically when I apply suggestions. If there's an option, I could disable it.

Ah okay, sorry ^^
nah, that's fine then :)

src/sensors/bno080sensor.cpp
TheDevMinerTV: I guess, yeah

deiteris: Updated indents as in example

src/sensors/bno080sensor.cpp
```diff
@@ -52,27 +52,23 @@ void BNO080Sensor::motionSetup()
...
-    bool useStabilization = false;
-#ifdef BNO_USE_ARVR_STABILIZATION
-    useStabilization = true;
+#if USE_6_AXIS
+    #if (IMU == IMU_BNO085 || IMU == IMU_BNO086) && BNO_USE_ARVR_STABILIZATION
+    imu.enableARVRStabilizedGameRotationVector(10);
+    #else
+    imu.enableGameRotationVector(10);
+    #endif
+    #if BNO_USE_MAGNETOMETER_CORRECTION
```
TheDevMinerTV, line 61:
```suggestion

    #if BNO_USE_MAGNETOMETER_CORRECTION
```

src/sensors/bno080sensor.cpp
```diff
@@ -115,55 +111,61 @@ void BNO080Sensor::motionLoop()
...
-#if ENABLE_INSPECTION
-                {
-                    Network::sendInspectionFusedIMUData(sensorId, quaternion);
-                }
-#endif
+            newMagData = true;
+        }
+    #endif // BNO_USE_MAGNETOMETER_CORRECTION
+#else // USE_6_AXIS
+        if (imu.hasNewQuat())
```
TheDevMinerTV, line 150:
```suggestion

        if (imu.hasNewQuat())
```

src/sensors/bno080sensor.cpp
```diff
@@ -213,19 +215,24 @@ void BNO080Sensor::sendData()
     {
         newData = false;
         Network::sendRotationData(&quaternion, DATA_TYPE_NORMAL, calibrationAccuracy, sensorId);
-        if (useMagnetometerAllTheTime)
-            Network::sendMagnetometerAccuracy(magneticAccuracyEstimate, sensorId);
+#if !USE_6_AXIS
```
TheDevMinerTV, line 218:
```suggestion

#if !USE_6_AXIS
```

TheDevMinerTV (changes requested):

TheDevMinerTV:
That should be it, but I can't test for function.

Eirenliel (approved):

PR #167, opened 2022-06-16
small adjustments for plattformio.ini

doormatt-dev:
updated the docs link and configured the serial monitor to display user typed text, this makes it easier to set wifi credentials if the server's wifi window doesn't work for whatever reason (it is still possible to type in the serial monitor without the "--echo" flag but the message will not be visible neither while typing nor after sending)


unlogisch04 (changes requested):
Change to 1 line. Adding a line more with a other option, or a option behind it, does not work.

platformio.ini
```diff
@@ -8,13 +8,15 @@
...
 ; See docs for configuration options and examples:
-; https://docs.slimevr.dev/configuring-project.html#configuring-platformioini
+; https://docs.slimevr.dev/firmware/configuring-project.html#1-configuring-platformioini
 ; ================================================
 
 [env]
 lib_deps=
   https://github.com/SlimeVR/CmdParser.git
 monitor_speed = 115200
+monitor_flags =
```
unlogisch04, line 18: the --echo can be on the same line as monitor_flags =

platformio.ini
```diff
@@ -8,13 +8,15 @@
...
+  --echo
```
unlogisch04, line 19:
```suggestion
monitor_flags = --echo
;monitor_rts = 0
;monitor_dtr = 0
```
Adding a line with a other flag did not work for me.
We can also add this monitor flags, so if someone wants to use they are there.

doormatt-dev: In line seems to work just as well as separate lines for me but if it's causing more flags not to register then i'll change that

In what cases would the rts and dtr lines be useful?

unlogisch04: It might not reboot the d1 mini on development board when you open the serial monitor if both are set to 0. Also the most ppl who will need echo will probably also need the rts / dtr, or they can use the WiFi window on slimevr server.

unlogisch04 (approved):

Eirenliel:
Why did you accept it with unresolved convo @unlogisch04? Idk what to do with it now :D

platformio.ini
doormatt-dev: i'll play around with those to hopefully reset the D1 when the serial monitor is opened, like the arduino IDE serial monitor does with most boards

unlogisch04: They only set the values when you open the serial monitor. It would be instead of using "pio device monitor --rts 0 --dtr 0".

TheDevMinerTV (changes requested):
What @unlogisch04 suggested would be cool to have

platformio.ini
doormatt-dev: seems like the CH340 on my D1s isn't able to auto reset the ESP when serial is opened, not even the Arduino serial monitor does it,
there should be a way somehow because it resets when flashing but i doubt there is a parameter that enables that

i'll add your changes though

doormatt-dev:
Woops didn't mean to re-request revirw from devMiner but I think the changes can be merged now and all requests were addressed

unlogisch04 (approved):

PR #168, opened 2022-06-21
Support for additional Sensors

mechbanana:
Added Support for some LSM303 Models

Used these repos heavily to get this working:

https://github.com/pololu/l3g-arduino
https://github.com/adafruit/Adafruit_LSM303DLH_Mag

I'm not entirely sure on the licensing on the adafruit version, this was really just a quick and dirty solution for me to get it up and running.


TheButlah:
BSD 3-clause: https://github.com/adafruit/Adafruit_LSM303DLH_Mag
MIT: https://github.com/pololu/l3g-arduino

Both of these are compatible, as long as you retain the original LICENSE disclaimers.

TheButlah:
Thanks for the contribution! Really excited about supporting these sensors.

I'd like to reduce the amount of code we need to maintain and allow us to update this code by trying to leave dependencies intact. In other words, I'd like to include the two libraries you used directly, and then have your modifications in separate files, or as "wrappers" on those libraries. Would that be possible? What is the nature of the changes you made?

mechbanana:
I reworked some bits to get the accelerometer/mag (LSM303s) and gyros (L3G) I'm not entirely sure why but using the Adafruit libraries only seemed to work for either the mag or accelerometer and never both.
I'll try fiddling with it a bit and get it up and running with just the imports

PR #173, opened 2022-06-30
Fix MPU-6500 cannot connect

inkuxuan:
some MPU-6500 seem to have their ID as 0x74.
This commit adds 0x74 as an legal deviceID for the MPU6050 library

### DIFF
--- a/lib/mpu6050/MPU6050.cpp
+++ b/lib/mpu6050/MPU6050.cpp
@@ -67,8 +67,12 @@ void MPU6050::initialize(uint8_t address) {
  */
 bool MPU6050::testConnection() {
     uint8_t deviceId = getDeviceID();
-    // some MPU-6500 seem to have their ID as 0x74
-    return deviceId == 0x68 || deviceId == 0x70 || deviceId == 0x71 || deviceId == 0x73 || deviceId == 0x74; // Allow any MPUs; 
+    // 0x68 -> MPU-6050
+    // 0x70 -> MPU-6500
+    // 0x71 -> MPU-9250
+    // 0x73 -> MPU-9255
+    // 0x74 -> MPU-6515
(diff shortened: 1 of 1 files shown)


deiteris:
They may report different IDs, but have you checked whether it works?

inkuxuan:
It was because a batch if MPU-6500s manufactored in 2021 won't work. And
immediately after I added 0x74 as a legal ID, they work fine.


Yury ***@***.***> 于 2022年6月30日周四 23:53写道：

unlogisch04:
0x74 should be a MPU6515 and not a MPU6500 but they might work with that code too

lib/mpu6050/MPU6050.cpp
```diff
@@ -67,7 +67,8 @@ void MPU6050::initialize(uint8_t address) {
  */
 bool MPU6050::testConnection() {
     uint8_t deviceId = getDeviceID();
-    return deviceId == 0x68 || deviceId == 0x70 || deviceId == 0x71 || deviceId == 0x73; // Allow any MPUs
+    // some MPU-6500 seem to have their ID as 0x74
+    return deviceId == 0x68 || deviceId == 0x70 || deviceId == 0x71 || deviceId == 0x73 || deviceId == 0x74; // Allow any MPUs;
```
unlogisch04, line 71: If we add any comment to 1 ID, i think we should name them all. Also we do no longer any MPUs
```suggestion
    // 0x68 -> MPU-6050
    // 0x70 -> MPU-6500
    // 0x71 -> MPU-9250
    // 0x73 -> MPU-9255
    // 0x74 -> MPU-6515
    return deviceId == 0x68 || deviceId == 0x70 || deviceId == 0x71 || deviceId == 0x73 || deviceId == 0x74; 
```

unlogisch04 (commented):
I suggest this changes.

unlogisch04 (approved):

Eirenliel (approved):

PR #178, opened 2022-07-11
ESP32-C3 integration

unlogisch04:
Draft Pull request for discussions and feedback's.

Todo:
- I2CScan need to be changed to work with ESP32C3. Missing here a Compiler flag to be sure that we have a ESP32C3.
- Board the firebeetle32c3 https://github.com/platformio/platform-espressif32/pull/829
- Testing on other boards.

Testing:
- [x] Testing ICM20948
- [x] Testing MPU6050
- Testing MPU9250 (i can not test that one)
- Testing BNO085 (i can not test that one)
- [x] Testing BMI160 (BMI160 seems to crash after some console "REBOOT" or after a short time the data seems only from magnetometer, not sure if it is only my BMI160 bad or a problem with the lib not correctly init.)
- [ ] Testing Server WiFi Window (does not work filter need to be changed to allow "Serielles USB-Gerät" )
-> does not work
- [x] Board dfrobot_beetle_esp32c3 (need atm custom board)
- [x] Board esp32-c3-devkitm-1
- [x] Board lolin_c3_mini
- [x] Board ESP8266
- [x] Board ESP32


unlogisch04:
With the firebeetle32c3 i can not connect with the WiFi Window to the server.
It reports as: "Serielles USB-Gerät (COM10)" on Windows. This will not trigger the filter of the Server.

TheButlah (commented):
Tested on my `ferrous_slime` esp32c3 board, and a d1 mini. Both with MPU6050. It worked, with the following minor tweaks:

- On the esp32c3, I lost most of the initial serial printouts done in the setup until I added a `delay(2000)` right after the serial was initialized. I think we should add this delay in on the esp32c3 build behind an `#ifdef`
- The firebeetle32c3 board doesn't exist, I assume @unlogisch04 must have added his own custom board description json. To get this mergeable, we should just have two options probably - the dev kit and the lolin_c3_mini. This whole esp32c3 section will then be commented out and users can pick which one.
- We should add a comment in the esp32c3 section of the `platformio.ini` that explains which board to pick - if your board uses USB <-> UART chip, pick the `esp32-c3-devkitm-1` board. If it uses the onboard USB serial directly (D+ and D- are connected to GPIO19 and GPIO18 directly), then use the `lolin_c3_mini`. If using the latter, you will need to push the Flash/boot button, or otherwise set the appropriate strapping pins. Link them to this: https://docs.espressif.com/projects/esptool/en/latest/esp32c3/advanced-topics/boot-mode-selection.html

unlogisch04:
I request to review and test. I did not squash till now, so changes can be tracked better.
@TheDevMinerTV @Eirenliel 
Is it ok to add 2 boards to the CI?

TheDevMinerTV:
we're not paying for CI minutes, so yes.
note to self: I have to parallelize the building at some point though.

unlogisch04:
It should be ready for review.
Only open point is, that the the Server: https://github.com/SlimeVR/SlimeVR-Server/issues/108#issuecomment-1208546510

TheButlah:
Will review maybe in a week when I get back home

README.md
```diff
@@ -48,6 +48,13 @@ Firmware can work with both ESP8266 and ESP32. Please edit `defines.h` and set y
   * LED will turn off when calibration is complete
   * You don't have to calibrate next time you power it off, calibration values will be saved for the next use
 
+## Infos about ESP32-C3 with direct connection to USB
+The ESP32-C3 has 2 ways to connect the serial port. One is directly via the onboard USB CDC or via the onboard UART.
+When the Chip is connected to the USB-CDC, the serial port shows is displayed as "USB Serial Port". The SlimeVR server is not currently connecting to this port.
+If you want to set your WiFi credentials, you can use the PlatformIO serial console.
+There you have to enter the following:
+`SET WIFI "SSID" "PASSWORD"`
```
TheDevMinerTV, line 56:
```suggestion
The ESP32-C3 has two ways to connect the serial port. One is directly via the onboard USB CDC or via the onboard UART.
When the chip is connected to the USB CDC, the serial port shows as `USB Serial Port` in Device Manager. The SlimeVR server will currently not connect to this port.
If you want to set your WiFi credentials, you can use the PlatformIO serial console.
There you have to enter the following: `SET WIFI "SSID" "PASSWORD"`
```

lib/i2cscan/i2cscan.cpp
```diff
@@ -1,11 +1,18 @@
 #include "i2cscan.h"
+#include "../../src/globals.h"
 
 #ifdef ESP8266
 uint8_t portArray[] = {16, 5, 4, 2, 14, 12, 13};
+uint8_t portExclude[] = {LED_PIN};
 String portMap[] = {"D0", "D1", "D2", "D4", "D5", "D6", "D7"};
+#elif defined(ESP32C3)  //ESP32C3 has not that much ports as a ESP32
```
TheDevMinerTV, line 8:
```suggestion
// ESP32C3 has not as many ports as the ESP32
#elif defined(ESP32C3)
```

lib/i2cscan/i2cscan.cpp
```diff
@@ -1,11 +1,18 @@
 #include "i2cscan.h"
+#include "../../src/globals.h"
```
TheDevMinerTV, line 2: Do you really need this include?

lib/i2cscan/i2cscan.cpp
```diff
@@ -43,12 +50,45 @@ namespace I2CSCAN
...
+        }
+        //Serial.printf("InArray: %d, %d, %d, result: %x\n", search, search2, size, ret);
+        return ret;
+    }
+    
     bool checkI2C(uint8_t i, uint8_t j)
     {
         bool found = false;
-        Wire.begin(portArray[i], portArray[j]);
+#if ESP32
```
TheDevMinerTV, line 88: Shouldn't this be for the ESP32C3?

lib/i2cscan/i2cscan.cpp
```diff
@@ -43,12 +50,45 @@ namespace I2CSCAN
         if(!found) {
             Serial.println("[ERR] I2C: No I2C devices found");
         }
+    // begin Reset the Wire interface back to the original Values
+    #if ESP32
+        Wire.end();
+    #endif
+        Wire.begin(static_cast<int>(PIN_IMU_SDA), static_cast<int>(PIN_IMU_SCL));
+    // end Reset the Wire interface back to the original Values
```
TheDevMinerTV, line 58:
```suggestion
#if ESP32
        Wire.end();
#endif

        // Reset the I2C interface back to it's original values
        Wire.begin(static_cast<int>(PIN_IMU_SDA), static_cast<int>(PIN_IMU_SCL));
```

lib/i2cscan/i2cscan.cpp
```diff
@@ -43,12 +50,45 @@ namespace I2CSCAN
...
+                if (search2 == arraysearch[k]) 
+                {
+                    ret = true;
+                    k = size;
+                }
+            }
+        }
+        //Serial.printf("InArray: %d, %d, %d, result: %x\n", search, search2, size, ret);
+        return ret;
+    }
```
TheDevMinerTV, line 83:
```suggestion
    boolean inArray(uint8_t value, uint8_t* array, size_t arraySize)
    {
        for (size_t i = 0; i < arraySize; i++)
        {
            if (value == array[i]) 
            {
                return true;
            }
        }

        return false;
    }
```

lib/i2cscan/i2cscan.cpp
```diff
@@ -33,7 +40,7 @@ namespace I2CSCAN
         {
             for (uint8_t j = 0; j < sizeof(portArray); j++)
             {
-                if (i != j)
+                if ((i != j) && !inArray(portArray[i], portArray[j], portExclude, sizeof(portExclude)))
```
TheDevMinerTV, line 43:
```suggestion
                if ((i != j) && !inArray(portArray[i], portExclude, sizeof(portExclude)) && !inArray(portArray[j], portExclude, sizeof(portExclude)))
```

lib/i2cscan/i2cscan.h
```diff
@@ -10,6 +10,7 @@ namespace I2CSCAN {
     bool isI2CExist(uint8_t addr);
     uint8_t pickDevice(uint8_t addr1, uint8_t addr2, bool scanIfNotFound);
     int clearBus(uint8_t SDA, uint8_t SCL);
+    boolean inArray(uint8_t search, uint8_t search2, uint8_t arraysearch[], uint8_t size);
```
TheDevMinerTV, line 13:
```suggestion
    boolean inArray(uint8_t value, uint8_t* arr, size_t arrSize);
```

TheDevMinerTV (changes requested):

ci/devices.json
```diff
@@ -5,13 +5,18 @@
     "board": "SLIMEVR"
   },
   {
-    "platform": "espressif32 @ 4.4.0",
+    "platform": "espressif32 @ 5.1.1",
```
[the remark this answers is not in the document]
unlogisch04: Forgot thx.

lib/i2cscan/i2cscan.cpp
unlogisch04: yes, it is needed for `uint8_t portExclude[] = {LED_PIN};`

lib/i2cscan/i2cscan.cpp
unlogisch04: No, it is for ESP32 (espressif32). They did add a check on the new versions (i think from 5.1 on)

lib/i2cscan/i2cscan.cpp
unlogisch04: Now we have to loop 2 times trough the same array. But agree it looks better.

ci/devices.json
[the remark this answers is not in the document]
unlogisch04: fixed with commit [ff879a8](https://github.com/SlimeVR/SlimeVR-Tracker-ESP/pull/178/commits/ff879a8dd88a51f2addd4c0ae69620154ed2fdb8)
also corrected not needed compile flags (new in board definition)

Eirenliel (approved):

lib/i2cscan/i2cscan.cpp
```diff
@@ -43,12 +51,34 @@ namespace I2CSCAN
...
         }
+#if ESP32
+        Wire.end();
+#endif
+
+        // Reset the I2C interface back to it's original values
+        Wire.begin(static_cast<int>(PIN_IMU_SDA), static_cast<int>(PIN_IMU_SCL));
     }
 
+    boolean inArray(uint8_t value, uint8_t* array, size_t arraySize)
```
TheDevMinerTV, line 62:
```suggestion
    bool inArray(uint8_t value, uint8_t* array, size_t arraySize)
```

lib/i2cscan/i2cscan.cpp
```diff
@@ -43,12 +51,34 @@ namespace I2CSCAN
...
+    }
+    
     bool checkI2C(uint8_t i, uint8_t j)
     {
         bool found = false;
-        Wire.begin(portArray[i], portArray[j]);
+#if ESP32
+        Wire.end();
+#endif
+        Wire.begin((int)portArray[i], (int)portArray[j]);
```
TheDevMinerTV, line 81:
```suggestion

#if ESP32
        Wire.end();
#endif

        Wire.begin((int)portArray[i], (int)portArray[j]);

```

lib/i2cscan/i2cscan.cpp
```diff
@@ -43,12 +51,34 @@ namespace I2CSCAN
         if(!found) {
             Serial.println("[ERR] I2C: No I2C devices found");
         }
+#if ESP32
+        Wire.end();
+#endif
```
TheDevMinerTV, line 56:
```suggestion

#if ESP32
        Wire.end();
#endif
```

ci/build.py
```diff
@@ -25,10 +27,19 @@ def __init__(self,  platform: str, board: Board, platformio_board: str) -> None:
...
+
+        if self.board == Board.LOLIN_C3_MINI:
+            section += "build_flags = \n"
+            section += " ${env.build_flags}\n"
+            section += " -DESP32C3\n"
+
+        if self.board == Board.ES32C3DEVKITM1:
+            section += "build_flags = \n"
+            section += " ${env.build_flags}\n"
+            section += " -DESP32C3\n"
```
TheDevMinerTV, line 42:
```suggestion
            section += " -DESP32C3\n"

```

TheDevMinerTV (changes requested):

TheDevMinerTV (approved):

PR #179, opened 2022-07-17
fixing error on ESP32 using BMI160

unlogisch04:
https://github.com/jrowberg/i2cdevlib/issues/670
https://github.com/espressif/arduino-esp32/issues/6674
https://github.com/espressif/arduino-esp32/issues/6674#issuecomment-1122606853


TheDevMinerTV:
doesn't this break all other ESPs and IMUs? can you test?

deiteris:
This shouldn't break IMUs, but we probably need to check on esp8266 since API behavior may be different.

unlogisch04:
Works on:
- [x] D1 Mini + BMI160
- [x] D1 Mini + MPU6050
- [x] D1 Mini + ICM20948
- [x] ESP32 + ICM20948
- [x] ESP32 + BMI160

PR #182, opened 2022-07-20
Scale accelerometer and magnetometer values

deiteris:
Adds scaling of accelerometer (to linear acceleration) and magnetometer (to microteslas) values to MPU9250 and BMI160. Mahony fusion probably should work better with scaled values instead of just calibrated sensor values.

Needs testing.

### DIFF
--- a/src/sensors/bmi160sensor.cpp
+++ b/src/sensors/bmi160sensor.cpp
@@ -25,8 +25,6 @@
 #include "network/network.h"
 #include "GlobalVars.h"
 
-#define SENSORS_GRAVITY_EARTH 9.80665
-
 // Typical sensitivity at 25C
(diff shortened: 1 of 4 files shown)


Eirenliel (approved):

src/sensors/bmi160sensor.cpp
```diff
@@ -25,13 +25,19 @@
 #include "network/network.h"
 #include "GlobalVars.h"
 
+#define SENSORS_GRAVITY_EARTH 9.80665
```
TheDevMinerTV, line 28: this should probably be in a different file so it's not duplicated

deiteris: Could move to `sensors.h`, I guess.

TheDevMinerTV: or to constants.h

PR #183, opened 2022-08-03
Adafruit NXP Precision Support

Rosdayle:
Relying solely on gyro is not recommended due to excess drift (Unsure if this is fixable)


Eirenliel (changes requested):
The code doesn't seem properly formatted, please format the code, thanks.

Loler920a:
Clarification on proper formatting?

TheDevMinerTV:
I'd prefer if you actually spelled out Adafruit in the sensor and class name.

src/configuration/CalibrationConfig.h
```diff
@@ -61,6 +61,19 @@ namespace SlimeVR {
...
+            float A_B[3];
+            float A_Ainv[3][3];
+
+            // magnetometer offsets and correction matrix
+            float M_B[3];
+            float M_Ainv[3][3];
+
+            // raw offsets, determined from gyro at rest
+            float G_off[3];
+        };
```
TheDevMinerTV, line 75: You can use the same calibration config as the MPU9250 as the fields are the same.

src/sensors/AdafruitNXPsensor.cpp
```diff
@@ -0,0 +1,291 @@
+/*
+    SlimeVR Code is placed under the MIT license
+    Copyright (c) 2021 Eiren Rain & SlimeVR contributors
```
TheDevMinerTV, line 3: Copyright invalid, should be 2022 \<your name\>

Loler920a:
Updated formatting

src/sensors/AdafruitNXPsensor.cpp
Loler920a: Updated

src/configuration/CalibrationConfig.h
Loler920a: ;(

src/sensors/AdafruitNXPsensor.h
```diff
@@ -0,0 +1,61 @@
+/*
+    SlimeVR Code is placed under the MIT license
+    Copyright (c) 2021 Eiren Rain & SlimeVR contributors
```
Eirenliel, line 3: Same copyright issue here

Eirenliel (changes requested):

Eirenliel:
New sensors will not be accepted unless part of Sfusion or has own internal fusion.

PR was very stale. If the support for this IMU is needed, please make a new one with sfusion.

PR #192, opened 2022-09-04
Make MPU9250 use the FIFO

kitlith:
Idea is that using the FIFO with give us more accurate information, if we're able to process all of the samples.

One thing that should be done is a way to signal visibly somewhere that the buffer has overrun, so that we can realize that that's actually happening when testing this.

I'm going to leave a review highlighting all the TODOs I left in the code, so that other people can remark, and decide if they need to be handled now or later.

### DIFF
--- a/src/sensors/mpu9250sensor.cpp
+++ b/src/sensors/mpu9250sensor.cpp
@@ -95,7 +95,7 @@ void MPU9250Sensor::motionSetup() {
     }
 
 #if not (defined(_MAHONY_H_) || defined(_MADGWICK_H_))
-    devStatus = imu.dmpInitialize();
+    uint8_t devStatus = imu.dmpInitialize();
     if(devStatus == 0){
         ledManager.pattern(50, 50, 5);
 
@@ -144,8 +144,9 @@ void MPU9250Sensor::motionLoop() {
     if(!dmpReady)
         return;
     Quaternion rawQuat{};
-    if(!imu.GetCurrentFIFOPacket(fifoBuffer,imu.dmpGetFIFOPacketSize())) return;
-    if(imu.dmpGetQuaternion(&rawQuat, fifoBuffer)) return; // FIFO CORRUPTED
+    uint8_t dmpPacket[packetSize];
+    if(!imu.GetCurrentFIFOPacket(dmpPacket, packetSize)) return;
+    if(imu.dmpGetQuaternion(&rawQuat, dmpPacket)) return; // FIFO CORRUPTED
     Quat quat(-rawQuat.y,rawQuat.x,rawQuat.z,rawQuat.w);
 
     getMPUScaled();
--- a/src/sensors/mpu9250sensor.cpp
+++ b/src/sensors/mpu9250sensor.cpp
@@ -356,8 +356,9 @@ void MPU9250Sensor::startCalibration(int calibrationType) {
         calibrationDataMag[i * 3 + 0] = my;
         calibrationDataMag[i * 3 + 1] = mx;
         calibrationDataMag[i * 3 + 2] = -mz;
-        Network::sendRawCalibrationData(calibrationDataAcc, CALIBRATION_TYPE_EXTERNAL_ACCEL, 0);
-        Network::sendRawCalibrationData(calibrationDataMag, CALIBRATION_TYPE_EXTERNAL_MAG, 0);
+        // Thought: make the server run magneto for us, don't store samples in memory?
+        Network::sendRawCalibrationData(calibrationDataAcc + i * 3, CALIBRATION_TYPE_EXTERNAL_ACCEL, 0);
+        Network::sendRawCalibrationData(calibrationDataMag + i * 3, CALIBRATION_TYPE_EXTERNAL_MAG, 0);
         ledManager.off();
         delay(250);
     }
--- a/src/sensors/mpu9250sensor.cpp
+++ b/src/sensors/mpu9250sensor.cpp
@@ -194,7 +194,7 @@ void MPU9250Sensor::motionLoop() {
         grav.y *= 2;
         grav.z *= 2;
 
-        this->imu.dmpGetAccel(&this->rawAccel, fifoBuffer);
+        this->imu.dmpGetAccel(&this->rawAccel, dmpPacket);
         this->imu.dmpGetLinearAccel(&this->rawAccel, &this->rawAccel, &grav);
 
         // convert acceleration to m/s^2 (implicitly casts to float)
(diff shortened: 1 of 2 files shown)


src/sensors/mpu9250sensor.cpp
```diff
@@ -116,12 +116,25 @@ void MPU9250Sensor::motionSetup() {
...
 #else
+    // NOTE: could probably combine these into less total writes, but this should work, and isn't time critical.
+    imu.setAccelFIFOEnabled(true);
+    imu.setXGyroFIFOEnabled(true);
+    imu.setYGyroFIFOEnabled(true);
+    imu.setZGyroFIFOEnabled(true);
+    imu.setSlave0FIFOEnabled(true);
+
+    // TODO: set a rate we prefer instead of getting the current rate from the device.
+    deltat = 1.0 / 1000.0 * (1 + imu.getRate());
```
kitlith, line 127: I could be wrong, but we don't appear to have ever been setting a specific rate anywhere? so I'm querying from the IMU what rate is set and computing timings based off of that, but we can change this to set it to a rate of our choosing.

src/sensors/mpu9250sensor.cpp
```diff
@@ -166,16 +182,29 @@ void MPU9250Sensor::motionLoop() {
...
 #else
-    unsigned long now = micros();
-    unsigned long deltat = now - last; //seconds since last update
-    last = now;
-    getMPUScaled();
+
+    union fifo_sample_raw buf;
+    uint16_t remaining_samples;
+    // TODO: would it be faster to read multiple samples at once
+    while (getNextSample (&buf, &remaining_samples)) {
```
kitlith, line 189: I started with a relatively complicated design that involved the ability to read multiple samples into a buffer at once, and then maybe focused too much on avoiding issues with strict aliasing, but settled on this simpler design in the end.

In essence, this TODO is asking the question "how much time is wasted doing individual i2c transactions instead of trying to batch them"? If the time is neglegible, then it's not important. If it's a lot, then maybe we should try adding this in.

might want to write a microbenchmark to figure this out, if we care enough to do so.

src/sensors/mpu9250sensor.cpp
```diff
@@ -166,16 +182,29 @@ void MPU9250Sensor::motionLoop() {
...
+        parseAccelData(buf.sample.accel);
+        parseGyroData(buf.sample.gyro);
+        parseMagData(buf.sample.mag);
+
+        // TODO: monitor magnetometer status
+        // buf.sample.mag_status;
+        // TODO: monitor interrupts
+        // imu.getIntStatus();
+        // TODO: monitor remaining_samples to ensure that the number is going down, not up.
+        // remaining_samples
```
kitlith, line 199: * actually looking at the magnetometer status would allow us to detect when the magnetometer reading is incorrect
 * looking at the interrupt status would allow us to handle/detect fifo overflow
 * monitoring remaining_samples would allow is to avoid an infinite loop scenario in the case that our sample processing loop is *slower* than the imu outputs data.

src/sensors/mpu9250sensor.cpp
```diff
@@ -382,4 +366,101 @@ void MPU9250Sensor::startCalibration(int calibrationType) {
...
+        #if useFullCalibrationMatrix == true
+            Axyz[i] = m_Calibration.A_Ainv[i][0] * temp[0] + m_Calibration.A_Ainv[i][1] * temp[1] + m_Calibration.A_Ainv[i][2] * temp[2];
+        #else
+            Axyz[i] = temp[i];
+        #endif
+    }
+}
+
+// TODO: refactor so that calibration/conversion to float is only done in one place.
+void MPU9250Sensor::parseGyroData(int16_t data[3]) {
```
kitlith, line 412: this TODO might be out of date. whoops.

src/sensors/mpu9250sensor.cpp
```diff
@@ -382,4 +366,101 @@ void MPU9250Sensor::startCalibration(int calibrationType) {
...
+    // reading big endian int16
+    Gxyz[0] = ((float)data[0] - m_Calibration.G_off[0]) * gscale; //250 LSB(d/s) default to radians/s
+    Gxyz[1] = ((float)data[1] - m_Calibration.G_off[1]) * gscale;
+    Gxyz[2] = ((float)data[2] - m_Calibration.G_off[2]) * gscale;
+}
+
+// really just an implementation detail of getNextSample...
+void MPU9250Sensor::swapFifoData(union fifo_sample_raw* sample) {
```
kitlith, line 420: I should consider moving the internals of swapFifoData directly into getNextSample

src/sensors/mpu9250sensor.cpp
```diff
@@ -382,4 +366,101 @@ void MPU9250Sensor::startCalibration(int calibrationType) {
...
+
+// thought experiments:
+// is a single burst i2c transaction faster than multiple? by how much?
+// how does that compare to the performance of a memcpy?
+// how does that compare to the performance of data fusion?
+// if we read an extra byte from the magnetometer (or otherwise did something funky)
+// we could read into a properly aligned array of fifo_samples (and not require a memcpy?)
+// TODO: strict aliasing might not be violated if we just read directly into a fifo_sample*.
+//  which means the union approach may be overcomplicated. *shrug*
+bool MPU9250Sensor::getNextSample(union fifo_sample_raw *buffer, uint16_t *remaining_count) {
```
kitlith, line 451: If it is overcomplicated, it should be in a way that doesn't affect performance.

src/sensors/mpu9250sensor.cpp
```diff
@@ -382,4 +366,101 @@ void MPU9250Sensor::startCalibration(int calibrationType) {
...
+    // compiler hint for the union, should be optimized away for optimization >= -O1 according to compiler explorer
+    memmove(&sample->sample, sample->raw, sensor_data_len);
+}
+
+// thought experiments:
+// is a single burst i2c transaction faster than multiple? by how much?
+// how does that compare to the performance of a memcpy?
+// how does that compare to the performance of data fusion?
+// if we read an extra byte from the magnetometer (or otherwise did something funky)
+// we could read into a properly aligned array of fifo_samples (and not require a memcpy?)
```
kitlith, line 448: This thought experiment is asking the same question as "would it be faster to read multiple samples at once" up above, except in more documentation and detail, in case this gets merged and someone is poking around for something to try.

The comment about an extra byte from the magnetometer is referring to the fact that a single sample is currently 19 bytes, because of the magnetometer status byte. If we had an array of `fifo_sample`, there would likely be a single byte of padding between each sample for 16 bit address alignment, because these structures mostly consist of 16 bit integers. If we read an extra byte from the magnetometer, then we could have a sample that is 20 bytes, and requires no additional padding bytes.

Eirenliel:
Please rebase to main.

Eirenliel (approved):

PR #200, opened 2022-10-06
Make magneto take a constant amount of memory, regardless of the number of samples.

### LINKED ISSUE #197: [Magneto] Handle infinite samples with constant memory usage
the first thing magneto does with the input sample is first expand each sample into a 10 item row/column, treat the input as a 10xN matrix, and multiply it by it's own transpose, resulting in a 10x10 matrix.

expanding out the math, it turns out that taking each sample as a 10x1 matrix, multiplying by its own transpose, then summing each resulting matrix together results in the same matrix, except for floating point imprecision. I'm not sure whether the result is more or less precise, but according to my testing it's fairly close.

https://blog.demofox.org/2017/01/02/incremental-least-squares-surface-and-hyper-volume-fitting/ suggests we can do it with even less space than a full 10x10 matrix due to the structure of the input, 19 elements instead of 100. nevertheless, even if we don't go that far, we can still save quite a bit of memory. (currently, for mpu9250 at least, we save 3\*300 elements, which gets expanded to 10\*300 elements, before we end up with the 10x10 matrix of 100 elements.)

However, since we'd be doing the calculations on the fly, we lose the current method of outlier rejection. If we wanted to keep outlier rejection, we'd need to find a way to do it 'online', as samples are streaming in.

kitlith:
Closes #197

Third commit is the real meat of this change, it is probably easier to review commit by commit due to the movement of the math functions of another file in the first commit.

I have tested this idea and observed negligible or no change in the calibration matrix given the same input.

I haven't changed anything about how the sensors use magneto, other than causing a net reduction in memory usage. I leave it up to other people to experiment with increasing the number of samples now that we're no longer memory constrained.


kitlith:
Once #201 is merged, the pull_request check should succeed.

TheDevMinerTV:
@kitlith you still have to rebase onto `main` to get the fix.

src/sensors/bmi160sensor.cpp
```diff
@@ -271,17 +271,16 @@ void BMI160Sensor::startCalibration(int calibrationType) {
...
     for (int i = 0; i < accelCalibrationSamples; i++)
     {
         ledManager.on();
 
         int16_t ax, ay, az;
         imu.getAcceleration(&ax, &ay, &az);
-        calibrationDataAcc[i * 3 + 0] = ax;
-        calibrationDataAcc[i * 3 + 1] = ay;
-        calibrationDataAcc[i * 3 + 2] = az;
+        magneto->sample(ax, ay, az);
```
TheDevMinerTV, line 283: Can we also use this class to calibrate the acceleration?

src/sensors/mpu9250sensor.cpp
```diff
@@ -279,23 +279,23 @@ void MPU9250Sensor::startCalibration(int calibrationType) {
...
     m_Logger.debug("Calculating calibration data...");
 
     float M_BAinv[4][3];
-    CalculateCalibration(calibrationDataMag, calibrationSamples, M_BAinv);
-    free(calibrationDataMag);
+    magneto->current_calibration(M_BAinv);
+    delete magneto;
 
     m_Logger.debug("[INFO] Magnetometer calibration matrix:");
     m_Logger.debug("{");
```
TheDevMinerTV, line 301: Below this (GitHub doesn't allow me to comment there), there's another code block that uses `getMotion9()`. Should that use `MagnetoCalibration` as well?

TheDevMinerTV (commented):
There's code for the ICM20498 missing

src/sensors/mpu9250sensor.cpp
kitlith: referring to the gyro calibration? No, it shouldn't.

Magneto takes a bunch of data points that should all lie on a sphere, and figures out how to transform them so they are actually on a sphere. This requires a constant vector that can move around arbitrarily. For accelerometer calibration, this is the gravity vector, and we "move" it around by rotating the tracker. For magnetometer calibration, this is the magnetic field, and we "move" it around by rotating the tracker.

We don't have something similar for a gyroscope currently, so instead we only try to compute the bias -- the average of a bunch of values.

Aside: The goal of this PR is simply to update the uses of magneto that already exist, since I had to change the interface for my improvement to magneto. Therefore, I only replaced the uses of `CalculateCalibration`, which was the old interface to magneto.

src/sensors/bmi160sensor.cpp
kitlith: Yes. I went into more detail in my other comment. Moreover, in my opinion for this PR, it's less of a matter of whether we *can* and more that we *already did*.

kitlith:
> There's code for the ICM20498 missing

The ICM20498 does not have any mode that currently uses magneto, and therefore is out of scope for this PR.

If someone writes something for the ICM20498 that doesn't use the built-in DMP, that would be the time to introduce the use of magneto.

TheDevMinerTV:
Ah okay, makes sense ^^

src/sensors/mpu9250sensor.cpp
TheDevMinerTV: Okay, makes sense

TheDevMinerTV:
Needs testing.

0forks:
This optimization has been in use in [my BMI branch](https://github.com/0forks/SlimeVR-Tracker-ESP-BMI160/tree/v3dev) for almost 2 months without any issues.

Eirenliel (approved):

PR #205, opened 2022-10-22
Fix BNO055 not working at all

Timocop:
BNO055s require external crystal use. Also, enable NDOF mode to use intelligent drift correction using mag.

### DIFF
--- a/src/sensors/bno055sensor.cpp
+++ b/src/sensors/bno055sensor.cpp
@@ -28,7 +28,7 @@
 void BNO055Sensor::motionSetup() {
     imu = Adafruit_BNO055(sensorId, addr);
     delay(3000);
-    if (!imu.begin(Adafruit_BNO055::OPERATION_MODE_IMUPLUS))
+    if (!imu.begin(Adafruit_BNO055::OPERATION_MODE_NDOF))
     {
         m_Logger.fatal("Can't connect to BNO055 at address 0x%02x", addr);
         ledManager.pattern(50, 50, 200);
@@ -39,6 +39,11 @@ void BNO055Sensor::motionSetup() {
     imu.setAxisRemap(Adafruit_BNO055::REMAP_CONFIG_P0);
     imu.setAxisSign(Adafruit_BNO055::REMAP_SIGN_P0);
     m_Logger.info("Connected to BNO055 at address 0x%02x", addr);
+
+    //Adafruit BNO055's use external crystal. Enable it, otherwise it does not work.
+    delay(1000);
+    imu.setExtCrystalUse(true);
+
     working = true;
     configured = true;
 }


src/sensors/bno055sensor.cpp
```diff
@@ -28,7 +28,7 @@
 void BNO055Sensor::motionSetup() {
     imu = Adafruit_BNO055(sensorId, addr);
     delay(3000);
-    if (!imu.begin(Adafruit_BNO055::OPERATION_MODE_IMUPLUS))
+    if (!imu.begin(Adafruit_BNO055::OPERATION_MODE_NDOF))
```
ButterscotchV, line 31: Using the magnetometer isn't always desirable, this could cause issues with people who have high magnetic interference. This setting could likely be tied to one of the magnetometer options in the debug.h file to allow for enabling/disabling it by user choice.

src/sensors/bno055sensor.cpp
```diff
@@ -39,6 +39,11 @@ void BNO055Sensor::motionSetup() {
     imu.setAxisRemap(Adafruit_BNO055::REMAP_CONFIG_P0);
     imu.setAxisSign(Adafruit_BNO055::REMAP_SIGN_P0);
     m_Logger.info("Connected to BNO055 at address 0x%02x", addr);
+
+    //Adafruit BNO055's use external crystal. Enable it, otherwise it does not work.
+    delay(1000);
+    imu.setExtCrystalUse(true);
```
ButterscotchV, line 45: Is there a case where the external crystal won't be present? For example, cheaper BNO055 breakout boards?

ButterscotchV (changes requested):

src/sensors/bno055sensor.cpp
Eirenliel: No, this is the correct fix. The crystal was disasbled because first board I made didn't have it, but all others do. I thought we had this fix for a while though? or true was even default.

deiteris: `imu.setExtCrystalUse(false);` previously was in the code, and we removed it in https://github.com/SlimeVR/SlimeVR-Tracker-ESP/commit/658db609e893617bb41dd8ebed3ca6cf46af9391. Now I checked the datasheet and realized that we actually didn't fix anything, because internal oscillator is used by default, lmao.
image

ButterscotchV: If there is an external crystal, it's probably better to use it, if all the breakout boards available use it then we could probably enable this

src/sensors/bno055sensor.cpp
Timocop: True! Forgot about that.

src/sensors/bno055sensor.cpp
unlogisch04: I did check the manual, we should use `imu.setExtCrystalUse(true);` if the board does not have one, or if it is faulty, it will switch back to internal.

https://www.bosch-sensortec.com/media/boschsensortec/downloads/datasheets/bst-bno055-ds000.pdf
PDF Page 110
> 5.5.1 External 32kHz Crystal Oscillator
> When switching to the external clock source, the BNO055 will check for integrity of the crystal’s
> signal. In case of improper signal (e.g. crystal damaged, broken/missing...), the BNO055 will
> automatically switch back to its internal clock source and clear the CLK_SEL bit in
> SYS_TRIGGER register.
>
> It takes minimum ~600ms to configure the external crystal and startup the BNO055.

ButterscotchV: Ah okay awesome, then this change is good 👍

src/sensors/bno055sensor.cpp
```diff
@@ -39,6 +43,11 @@ void BNO055Sensor::motionSetup() {
     imu.setAxisRemap(Adafruit_BNO055::REMAP_CONFIG_P0);
     imu.setAxisSign(Adafruit_BNO055::REMAP_SIGN_P0);
     m_Logger.info("Connected to BNO055 at address 0x%02x", addr);
+
+    //Adafruit BNO055's use external crystal. Enable it, otherwise it does not work.
+    delay(1000);
```
Eirenliel, line 48: Why is there a delay of 1000? I didn't notice that we need it during testing.

Eirenliel (changes requested):

src/sensors/bno055sensor.cpp
Timocop: Removed the double delay. But the 1000ms delay should still be present to configure the external crystal (which is like ~600ms).

Eirenliel (approved):

PR #220, opened 2023-01-20
Improved BMI160 support

0forks:
- Implements proper data sampling.
- Adds BMI160-specific defines, experimental magnetometer support, axis remapping, new accelerometer calibration method, initial temperature calibration support.
- Limits send rate to 120hz for BMI160.

This is not the final version, and there will be breaking backwards compatibility changes due to lack of proper calibration config serialization, but consider merging this anyway - the current BMI160 implementation has many flaws.

This should be a minor version bump, as major version 0 implies initial development phase and should not be considered stable anyway.

Tested on ESP8266 and ESP32.
Needs basic testing with other IMUs.


Eirenliel:
i want some ppl to review it and test if it doesn't break BNO085

src/GlobalVars.h
```diff
@@ -30,6 +30,7 @@
 #include "sensors/SensorManager.h"
 
 extern SlimeVR::LEDManager ledManager;
+extern SlimeVR::Sensors::SensorManager sensorManager;
```
Eirenliel, line 33: duplicate!

src/defines_bmi160.h
```diff
@@ -0,0 +1,73 @@
+/*
+    SlimeVR Code is placed under the MIT license
+    Copyright (c) 2021 Eiren Rain
```
Eirenliel, line 3: Copyright here and in other new files should be `Copyright (c) 2022 SlimeVR Contributors`

src/defines_sensitivity.h
```diff
@@ -0,0 +1,60 @@
+/*
+    SlimeVR Code is placed under the MIT license
+    Copyright (c) 2021 Eiren Rain
```
Eirenliel, line 3: Same as above.

src/motionprocessing/GyroTemperatureCalibrator.cpp
```diff
@@ -0,0 +1,207 @@
+/*
+    SlimeVR Code is placed under the MIT license
+    Copyright (c) 2021 Eiren Rain & SlimeVR contributors
```
Eirenliel, line 3: Same

src/motionprocessing/GyroTemperatureCalibrator.h
```diff
@@ -0,0 +1,223 @@
+/*
+    SlimeVR Code is placed under the MIT license
+    Copyright (c) 2021 Eiren Rain & SlimeVR contributors
```
Eirenliel, line 3: Same

src/motionprocessing/OnlinePolyfit.h
```diff
@@ -0,0 +1,125 @@
+/*
+    SlimeVR Code is placed under the MIT license
+    Copyright (c) 2021 Eiren Rain & SlimeVR contributors
```
Eirenliel, line 3: Same

Eirenliel (changes requested):
Actually @TheButlah what license should we put at the top of the files now?

src/configuration/CalibrationConfig.h
```diff
@@ -33,6 +33,10 @@ namespace SlimeVR {
             float A_B[3];
             float A_Ainv[3][3];
 
+            // magnetometer offsets and correction matrix
+            float M_B[3];
+            float M_Ainv[3][3];
```
TheDevMinerTV, line 38: You should implement a version upgrader for this.

TheDevMinerTV (changes requested):

src/sensors/bmi160sensor.cpp
```diff
@@ -110,110 +140,525 @@ void BMI160Sensor::motionSetup() {
...
+        #if BMI160_USE_VQF
+            vqf.updateGyr(Gxyz, (double)dtMicros * 1.0e-6);
+        #else
+            mahony.updateInto(qwxyz, Axyz[0], Axyz[1], Axyz[2], Gxyz[0], Gxyz[1], Gxyz[2], ((float)dtMicros) * 1.0e-6);
+        #endif
+    #else
+        #if BMI160_USE_VQF
+            vqf.updateGyr(Gxyz, (double)dtMicros * 1.0e-6);
+        #else
+            mahonyQuaternionUpdate(qwxyz, Axyz[0], Axyz[1], Axyz[2], Gxyz[0], Gxyz[1], Gxyz[2], Mxyz[0], Mxyz[1], Mxyz[2], ((float)dtMicros) * 1.0e-6f);
```
deiteris, line 558: At this point, we have 3 versions of mahony fusion. Are you sure you need C++ and C version side by side?

0forks: There's a problem with multiple instances of pure C version due to use of static global variables - it might cause weird behavior with aux trackers, ideally it should be replaced with the C++ class everywhere.

src/configuration/Configuration.cpp
```diff
@@ -156,34 +159,34 @@ namespace SlimeVR {
...
 
-                    calibrations = LittleFS.open("/calibrations");
+                    calibrations = LittleFS.open(DIR_CALIBRATIONS);
                 }
 
                 if (!calibrations.isDirectory()) {
                     calibrations.close();
 
-                    m_Logger.warn("Found file instead of directory: /calibrations");
+                    m_Logger.warn("Found file instead of directory: " DIR_CALIBRATIONS);
```
TheButlah, line 177: @unlogisch04 this is an implicit string literal concatenation - it should work fine, if you were concerned.
https://www.techiedelight.com/concatenate-string-literals-c-cpp/

src/serial/serialcommands.cpp
```diff
@@ -192,12 +192,41 @@ namespace SerialCommands {
...
+                sensorManager.getSecond()->printDebugTemperatureCalibrationState();
+                return;
+            } else if (parser->equalCmdParam(1, "RESET")) {
+                sensorManager.getFirst()->resetTemperatureCalibrationState();
+                sensorManager.getSecond()->resetTemperatureCalibrationState();
+                return;
+            } else if (parser->equalCmdParam(1, "SAVE")) {
+                sensorManager.getFirst()->saveTemperatureCalibration();
+                sensorManager.getSecond()->saveTemperatureCalibration();
+                return;
```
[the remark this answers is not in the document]
0forks: It needs to fall through when passing and unknown command as an argument, so I found a nested condition to be more compact.

lib/bmi160/BMI160.h
```diff
@@ -464,10 +544,71 @@ typedef enum {
...
+#define QMC_CFG_ODR_10HZ        0b00 << 2
+#define QMC_CFG_ODR_50HZ        0b01 << 2
+#define QMC_CFG_ODR_100HZ       0b10 << 2
+#define QMC_CFG_ODR_200HZ       0b11 << 2
+#define QMC_CFG_RNG_2G          0b00 << 4
+#define QMC_CFG_RNG_8G          0b01 << 4
+#define QMC_CFG_OSR_512         0b00 << 6
+#define QMC_CFG_OSR_256         0b01 << 6
+#define QMC_CFG_OSR_128         0b10 << 6
+#define QMC_CFG_OSR_64          0b11 << 6
```
unlogisch04, line 598: QMC and HMC is not used in BMI160 directly.
Move it to 2 new files, 1 for QMC and 1 for HMC.

src/configuration/Configuration.cpp
unlogisch04: thx

Eirenliel:
> > Actually @TheButlah what license should we put at the top of the files now?
> 
> You should just remove all of the license blurbs in the files. License blurbs are dumb, and aren't actually necessary. It only creates unecessary noise in the code.

they're useful when the file is copied from somewhere

TheButlah:
> they're useful when the file is copied from somewhere

If the file is copied from somewhere, thats a problem. Should not do that. Its no longer clear now which file is licensed under slimevr and which is licensed under another license.

I suggest that if a file is copied. Mark it as such and put a link to the *original* license in that file. but for slimevr authored files, don't do that.

Eirenliel (approved):

Eirenliel (approved):

TheButlah:
@0forks can you summarize in this PR which if any third party libraries have been added, whether they were modified, and link to the commit that the code was from?

0forks:
So, currently the third party code that was added/modified is the following:

- VQF v2: https://github.com/dlaidig/vqf/commit/1f76231acd6410a1657c787d028779b1a14fc2fb
  Modified `updateGyr(const vqf_real_t gyr[3], double gyrTs)` to add `gyrTs` time delta) to support gyro timestamps, everything else is assumed to be coming at fixed rate, as recomputing filtering coefficients for accel/mag would be very slow, and it seems to perform decently regardless.

- Added a separate rest detection class for use during accel calibration and fusion where rest detection isn't provided:
https://github.com/SlimeVR/SlimeVR-Tracker-ESP/blob/4cd8cfe891374e9d15031b3f5eec6cd6394a795d/src/motionprocessing/RestDetection.h#L1-L10

- Some fusion code was modified to compile with C++17:
  The `beta` variable in Madgwick interferes with something global, needed simple renaming)
  https://github.com/SlimeVR/SlimeVR-Tracker-ESP/blob/4cd8cfe891374e9d15031b3f5eec6cd6394a795d/lib/magneto/madgwick.cpp#L3
- Mahony fusion uses static global variables, which won't work for multiple instances:
    https://github.com/SlimeVR/SlimeVR-Tracker-ESP/blob/4cd8cfe891374e9d15031b3f5eec6cd6394a795d/lib/magneto/mahony.cpp#L32
  Currently only 6DOF Mahony is moved to a separate class:
    https://github.com/SlimeVR/SlimeVR-Tracker-ESP/blob/4cd8cfe891374e9d15031b3f5eec6cd6394a795d/lib/magneto/mahony.h#L32-L38
 - Some additions were made to BMI160 library, particularly magnetometer support, timestamp readout, FIFO header definitions, error handling. The library needs to be completely rewritten though.

I do realize that there are many code style and consistency issues with library code here, perhaps fixes could be done in another PR, if that is okay, as this one is already huge.

Maybe we should remove other fusion code in favor of (Basic)VQF. And as it will require modifying code for other IMUs and testing on hardware I don't have, I also think this goes slightly out of scope for this PR.

src/defines_sensitivity.h
```diff
@@ -0,0 +1,60 @@
...
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+    OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+    THE SOFTWARE.
+*/
+
+#ifndef DEFINES_SENSITIVITY_H
+#define DEFINES_SENSITIVITY_H
+
+// ==========================
+// Sensitivity calibration
```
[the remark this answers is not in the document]
0forks: It doesn't seem possible to compute a full calibration matrix for gyro on device, it simply doesn't have enough RAM and computing power. TCP was unusable when I tested on 8266. Current sensitivity calibration implementation is experimental and somewhat hidden for an average user. For now it is what it is, unfortunately.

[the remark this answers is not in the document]
Eirenliel: I want us to outline future firmware code improvements later, and this definetely should be there.

RexaOnGit:
> I do realize that there are many code style and consistency issues with library code here, perhaps fixes could be done in another PR, if that is okay, as this one is already huge.

I'm working on a fork based on v3dev that makes IMU count scalable and adds an SPI interface option for BMI160. I found that the bmi160 library was copied from a source with incompatible license, so I had to find a new library and fork that.

This requires at least some refactoring in multiple files. We could address the style issues there in some way.

0forks:
> IMU count scalable and adds an SPI

You may run into fusion scalability issues beyond 4 IMUs on 8266 as it doesn't have an FPU, I hope you have an ESP32 or something else with an FPU.

RexaOnGit:
> You may run into fusion scalability issues beyond 4 IMUs on 8266 as it doesn't have an FPU, I hope you have an ESP32 or something else with an FPU.

I do have esp8266, but I don't expect to need more than 4 IMUs connected. Maximums can be set for each combination of mcu, imu, and interface to prevent overloading hardware.

TheButlah:
> perhaps fixes could be done in another PR, if that is okay, as this one is already huge.

Agreed. 

> I found that the bmi160 library was copied from a source with incompatible license,

To confirm, @0forks you did not add any non-mit licensed code in this PR right? The incompatible license rexa is referring to predated this PR if I understand correctly?

RexaOnGit:
> To confirm, 0forks you did not add any viral licensed code in this PR right (LGPL, GPL, etc)? The incompatible license rexa is referring to predated this PR if I understand correctly?

The addition of BMI160 library associated with an LGPL predated this by PR by about 2 years based on the records I can see. I of course can't speak for anything 0 added.

0forks:
> you did not add any viral licensed code in this PR

VQF is MIT licensed, and the BMI library was already in the main branch. I think I did attribute all third party code I've added correctly.

Though I also did use some code from mag branches as reference, when adding mag support: https://github.com/deiteris/SlimeVR-Tracker-ESP/blob/hmc-mag/lib/bmi160/BMI160.cpp#L115

TheButlah (approved):
I did not really review this PR, but it has been getting stale and also has been tested for use by many. Advocating for a merge, with subsequent cleanup.

PR #222, opened 2023-01-22
BNO08x Calibration changes

Timocop:
Currently, BNO08x does only calibrate gyro and relies on factory accelerator and magnetometer calibration.
This causes a bunch of problems when using BNO08x with an enabled magnetometer (heavy uncorrectable drift).

Additionally, instead of running the BNO in calibration mode all the time which can cause garbage-in, and garbage-out results, this request includes a strict calibration method. This method is basically the same as an MPU9250.
The device will start calibration when it's on a stable surface and end calibration mode in - predefined - 30 seconds.
This gives the user enough time to calibrate gyro and magnetometer.

This has been tested - with an enabled magnetometer - successfully for one month with very good and stable results.


Eirenliel:
Hi. Thanks for the PR, we wanted to stop calibration of BNO after few seconds for a while. That said I have a few problems:

1) It explicitly enables gyro calibration, do we need it? Self-calibration is enabled by default. I was planning to just disable self-calibration after like 15 seconds from the start. Why I think it's better is because currently nothing actually breaks if user start trackers on their body and not do any calibration, the results might be worse, but nothing breaks too much. I don't know if this behaviour will continue if we explicitly call calibration?
2) 30 seconds is way too long for gyro calibration! It actually needs less than 5 seconds to calibrate zero offset.
3) Since it affects production slimes, we need to give it to way more people to test before it can be merged.

carl-anders:
> Why I think it's better is because currently nothing actually breaks if user start trackers on their body and not do any calibration, the results might be worse, but nothing breaks too much.

This is not that applicable to this PR, but I have some thoughts about this:
I think it should be a toggleable option of sorts. It's worth trying out a few configurations to see what works best for normal slime users. But it might also be worth it to have a setting (or more) for this that can be changed from the server. Where the server UI would show the advantages / disadvantages of turning the setting on. The configuration would be sent to the tracker without a need for flashing new firmware, similar to how wifi settings is saved, preferably sent with the normal firmware UDP method instead of serial only.

Timocop:
> 30 seconds is way too long for gyro calibration! It actually needs less than 5 seconds to calibrate zero offset.

``BNO_SELF_CALIBRATION_TIME`` is just the cut-off time to end calibration and save. (BNO saves every 5 minutes by default)
The result doesn't get worse if the user picks up the device at that time and uses it (same with magnetometer calibration). I've added a minimum time of 10 seconds to give the BNO time to calibrate the gyro just in case. Usually, gyro calibration takes a while to calculate drift bias, I'm not sure if the BNO does it 1:1 like that too. Just to be sure it's at least 10 seconds.

 > Why I think it's better is because currently nothing actually breaks if user start trackers on their body and not do any calibration

The calibration process is skipped when the user moves the device (e.g putting on their body) and relies on the last saved calibration instead. I doubt you need to re-calibrate the gyro everytime?

Eirenliel:
okay so does calling `imu.calibrateGyro();` changes anything from the default state?

Timocop:
Default state is factory calibration stored in DCD if ``imu.calibrateGyro();`` has never been called ever. Just like the accelerometer and magnetometer. Unless you send the DCD clear command manually to clear the flash (which I have done and tested too).
Dynamic calibration is not enabled by default. As stated in the docs from CEVA that running dynamic calibration in general use is discouraged.

https://www.ceva-dsp.com/wp-content/uploads/2019/09/BNO080-BNO085-Sesnor-Calibration-Procedure.pdf

> During normal AR/VR application use, Hillcrest recommends that dynamic calibration for the gyroscope and
magnetometer be disabled to avoid new calibrations being created during gameplay which can
cause unwanted motion artifacts

Eirenliel:
i thought it's enabled by deault? am i misremembering? you also said:

> Additionally, instead of running the BNO in calibration mode all the time which can cause garbage-in, and garbage-out results

i assumed that's the state on boot

Timocop:
Oh, no, that's when dynamic calibration is running continuously. Gyro and magnetometer (and probably accelerometer too) calibration will slowly "malform" over time. ICM20948 suffers from the same problem.

Eirenliel:
So the recomendation to start slimes on flat stable surgace is wrong for BNO085, are you sure? :P

So since we don't have to calibrate every time on boot, I think we should change how calibration is started to how it was always planned - via user command from the server, not automatically boot.

Timocop:
> So the recomendation to start slimes on flat stable surface is wrong for BNO085

With this pull request, yes. You only have to do it once, ever. After the first full calibration, it will use the last saved DCD calibration if you skip the calibration process.
This works perfectly for gyro and accelerometer, but the magnetometer needs some special care. Depending how the environment changes you might need to re-calibrate the magnetometer.

# Calibration Demo
[https://gfycat.com/WhoppingCooperativeArcticwolf](https://gfycat.com/WhoppingCooperativeArcticwolf)

# Skip Calibration Demo
[https://gfycat.com/DampYellowIndianrockpython](https://gfycat.com/DampYellowIndianrockpython)

Eirenliel:
> With this pull request, yes. 

I'm not talking about this PR, we're recommending people to do it currently, but if calibration is not enabled on boot, we don't need to.

Eirenliel:
Since it's not recommended to use magnetometer and it's not used by default, I think we should not calibrate on boot, unless user explicitly wants to recalibrate by a command from the server.

Eirenliel:
Sorry for being picky @Timocop, BNO code affects production so we take any changes to it very seriously.

Timocop:
No worries, all good. I know it's a critical area. Just sharing my experience and improvements as best as I can.

Eirenliel:
@Timocop do you have any comments on my other comments?

> I'm not talking about this PR, we're recommending people to do it currently, but if calibration is not enabled on boot, we don't need to.

> Since it's not recommended to use magnetometer and it's not used by default, I think we should not calibrate on boot, unless user explicitly wants to recalibrate by a command from the server.

I also would love this and #223 to be merged in a beta branch and given to people to test.

Eirenliel:
Here's the reply from CEVA:

The BNO085 enables the following dynamic calibration by default. If you need any other calibration (I do not see any reason for your application), you need to enable. Otherwise, you do not need to do anything.

In SHTP mode:

* Accelerometer Zero gravity offset (ZGO) calibration in 3 dimensional (SH2_CAL_ACCEL)
* Magnetometer hard and soft iron effect calibration (SH2_CAL_MAG)
* Gyroscope Zero rate offset (ZRO) calibration, on-table based (SH2_CAL_ON_TABLE, please note this can’t be disabled and always enabled). SH2_CAL_ON_TABLE is more accurate than in-hand calibration, SH2_CAL_GYRO (by default disabled).

Eirenliel:
Another reply:

SH2_CAL_ON_TABLE can’t be disabled and always enabled. Even though we mention that flag, there is no internal API to disable it.

If your device can be placed on table, this is the best option to calibrate the gyroscope ZRO. If your device can’t, then you may need to enable SH2_CAL_GYRO (in-hand based). This is less accurate (looking for stable state instead of stationary state) for ZRO calibration.

You can suggest to place your device on-table or enable sensor in the charging station so the BNO08X can learn the ZRO (on-tabled based) with the temperature compensation.

Eirenliel:
So I talked with CEVA engineer. They don't recommend doing what this PR does, this enables SH2_CAL_GYRO (in-hand calibration) which is worse than SH2_CAL_ON_TABLE that can't actually be disabled. So just placing it on the table will activate on-table calibration automatically no matter if it's after boot or later. Mag calibration is also always enabled, you don't need to call for it, just do 8-motion and it will calibrate hard and soft iron offsets automatically.

Timocop:
I think you might misunderstand what this PR does (maybe I didn't explain it well enough lol). 

Even without this PR, the current SlimeVR BNO library already calls ``calibrateGyro()`` and does in-hand calibration, *forever*. 
So you are basically already using the worst possible method that you told me about.

This PR does not change the current behavior from the current SlimeVR ESP, instead, it adds some indicators to guide calibration and enable accelerometer and magnetometer calibration.
What this PR actually does:
 - **Tell the user if the device is stable/on-table and ready for calibration.**
   - If the device is stable/on-table on boot, the device will go in calibration mode.
   - If the device is moving on boot, calibration will be skipped and load previous DCD calibration instead.
   - (This seems to be only useful for accelerometer and magnetometer)
 - **Enable accelerometer and magnetometer calibration because its factory and disabled by default.**
(Despite what the CEVA engineer says, accelerometer and magnetometer calibration is not on by default, you can test it yourself. It would also conflict with what the BNO calibration paper says too)
 - **Stop all dynamic calibration at a given time to avoid garbage readings.**

Also, as the Video above shows, the BNO08x takes its time to detect if it's stable/on-table (~10 seconds). So your documentation about "laying it on the ground for 3 seconds to calibrate gyro" might be wrong too. Since there is no indicator of when the device is ready or not and the user could pick the device up before the BNO could even attempt to calibrate the gyro. 

This PR basically adds an indicator to avoid user error using ``getStabilityClassifier()`` to tell if the device is stable/on-table and then starts calibration. If the device is moving, it will use the last saved DCD calibration and skip calibration altogether. It *might* be true that gyro will calibrate in runtime when it detects itself being in stable/on-table classifier but it's never mentioned in the BNO calibration document nor have I seen a comment about it in the code. As I see it, there is absolutely no reason why you would re-calibrate gyro every time and might only calibrate gyro when the DCD is clear. It's really hard to test. 

But I can already tell you, the accelerometer and magnetometer dynamic calibration **are not enabled by default** so I hardly doubt any other calibration methods are enabled by default as well. This is why I created this in the first place. The Magnetometer would *never** calibrate and would always use factory-saved DCD calibration no matter how many times I orientated the device. Enabling magnetometer calibration with ``calibrateAll()`` (or ``calibrateMag()``) did work.

There could be library differences as even the Adafruit BNO080 examples say that you should have ``calibrateAll()`` running at all times. 
https://github.com/sparkfun/SparkFun_BNO080_Arduino_Library/blob/main/examples/Example9-Calibrate/Example9-Calibrate.ino#L41
But in my tests, thats a very very bad idea.

Timocop:
Im using the Adafruit BNO085 breakout board btw.

Actually, i guess it depends on what RotationVector you use.
https://discourse.agopengps.com/t/bno085-or-bno080-imu/2993/72
https://discourse.agopengps.com/t/bno085-or-bno080-imu/2993/74

But then again, I already tested all rotation vectors. I guess I will do some more testing.
Since this algorithm hell is not very well documented.

Eirenliel:
> Even without this PR, the current SlimeVR BNO library already calls calibrateGyro() and does in-hand calibration, forever.
So you are basically already using the worst possible method that you told me about.

It's not called anywhere, @Timocop.

Timocop:
https://github.com/SlimeVR/SlimeVR-Tracker-ESP/blob/main/src/sensors/bno080sensor.cpp#L255

TheDevMinerTV:
That function is never called.

Timocop:
Wow, I really did not see the very obvious dead code. My forehead is kissing my desktop table right now...

But anyways, having an indicator for when the device is stable/on-table is probably a good idea. Since there's not much of a visual feedback for it.
And of course, mag calibration, because for some reason without ``calibrateAll()/calibrateMag()`` it would not calibrate in runtime. And stop calibration too so it would not break it either and have an ICM moment.

I will probably update this on my fork and test around further.

PR #221, opened 2023-01-20
Add ClockStretchLimit/Timeout to ESP32 for BNO085

unlogisch04:
Need to be tested.

PR #190, opened 2022-08-13
Implemented support for FSP201 by CEVA

Eirenliel:
Experimental support for engineering sample of FSP201 IMU/MCU by CEVA.

There is a bit of weird defines branching in main due to either bug in engineering samples of specifics of FSP201, so for now i hardcoded it. Waiting from CEVA on more info if this is intended or bug.

PR #223, opened 2023-01-23
Adjusted startup sequence of BNO08X sensors to be more flexible regar…

feldhausenryan:
Adjusted startup sequence of BNO08X sensors to be more flexible regarding packet timing/ordering for reset and productid request/response messages. This replaces inconsistent delay()s through the initialization process of the sensor and should result in a more reliable initial connection to the BNO08X sensors.

### DIFF
--- a/lib/bno080/BNO080.cpp
+++ b/lib/bno080/BNO080.cpp
@@ -66,12 +66,16 @@ boolean BNO080::begin(uint8_t deviceAddress, TwoWire &wirePort, uint8_t intPin)
 
 	//Transmit packet on channel 2, 2 bytes
 	sendPacket(CHANNEL_CONTROL, 2);
+	_debugPort->println(F("GetBoardInfoSent"));
 
-	//Now we wait for response
-	if (receivePacket() == true)
-	{
-		if (shtpData[0] == SHTP_REPORT_PRODUCT_ID_RESPONSE)
-		{
+	uint32_t tInitialResetTimeMS = millis();
+	bool tBoardInfoReceived = false;
+
+	// Wait max 2.5s for the product_id_response and ignore other packets received during that time.
+	while (millis() - tInitialResetTimeMS < 2500 && (!tBoardInfoReceived)) {
+		receivePacket();
+		if (shtpHeader[2] == 2 && shtpData[0] == SHTP_REPORT_PRODUCT_ID_RESPONSE) {
+			tBoardInfoReceived = true;
 			swMajor = shtpData[2];
 			swMinor = shtpData[3];
 			swPartNumber = ((uint32_t)shtpData[7] << 24) | ((uint32_t)shtpData[6] << 16) | ((uint32_t)shtpData[5] << 8) | ((uint32_t)shtpData[4]);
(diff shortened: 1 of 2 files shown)


feldhausenryan:
Context for this PR is conversation inside ESP32-C3 + BNO08X thread in the dev-forum section on discord.

lib/bno080/BNO080.cpp
```diff
@@ -1038,13 +1068,8 @@ void BNO080::softReset(void)
...
 
-	//Read all incoming data and flush it
-	delay(50);
-	while (receivePacket() == true)
-		; //delay(1);
-	delay(50);
-	while (receivePacket() == true)
-		; //delay(1);
+	waitForCompletedReset();
+	waitForCompletedReset();
```
Eirenliel, line 1072: Do we need to do it twice after the change? I think it was doing this for both packets to arrive.

Eirenliel (changes requested):

lib/bno080/BNO080.cpp
feldhausenryan: The BNO resets twice on reset. So it sends both messages twice for a total of 4 messages. That said I do not want this merged yet without some extra testing from people who have the same hardware combination.

Eirenliel: I see, thank you.

TheButlah:
I will test this tn on my d1 mini, my ferrous_slime v0.1.0 (no i2c pullups), ferrous_slime v0.4.0 (10kohm i2c pullups), and seeed studio xiao esp32c3. I will test in combination with adafruit bno080 and aliexpress bno080.

TheButlah:
* ferrous_slime v0.4.0:  adafruit bno and aliexpres bno both worked, but the old firmware already worked too on that board. I didn't test these super rigourously, maybe only like twice on each board.
* xiao esp32c3 & aliexpress bno: Failed 3/10 times I tried with: [ERR] I2C: Can't find I2C device on provided addresses, scanning for all I2C devices and returning.
* xiao esp32c3 & adafruit bno: failed 2/10 times with the same error

TheButlah:
@feldhausenryan pointed out in discord that the 
```
[ERR] I2C: Can't find I2C device on provided addresses, scanning for all I2C devices and returning.
```
error happens in i2cscan which hasn't been modified in this PR. So the numbers above are probably just random luck.

TheButlah:
The above i2cscan error is fixed when combined with #221. This pr fixes the 
> [FATAL] [BNO080Sensor:0] Can't connect to BNO085 at address 0x4a

So both PRs should be merged, imo.

Please note that I have not really looked too much at the *implementation* of the code in this PR. I merely tested the results.

TheButlah (approved):

TheButlah:
BTW, @feldhausenryan it would be cool if you submit this to the main adafruit repo as well. If you have time.

Eirenliel (approved):
I tried to change boot sequence in #190 but you did it properly, gj thanks

unlogisch04:
@TheButlah i belive it is the SparkFun Library
https://github.com/sparkfun/SparkFun_BNO080_Arduino_Library/blob/main/src/SparkFun_BNO080_Arduino_Library.cpp
They still have the https://github.com/sparkfun/SparkFun_BNO080_Arduino_Library/issues/72 issue open. Maybe somewhere there.

Eirenliel:
It brokey... https://github.com/SlimeVR/SlimeVR-Tracker-ESP/issues/264

PR #225, opened 2023-01-24
Proper battery sense fix

Vyolex:
This commit properly "fixes" the battery sense, in the previous fix I submitted (#139) I simply swapped the R1 and R2 values not noticing that the formula itself was wrong, not the default values. 

This rectifies that mistake by swapping the values again and fixing the formula instead. Now it should be consistent with the comments and the actual values inside the datasheet again and should cause less confusion.

This fix is OPTIONAL as it does not change anything relating to the behavior and only tries to improve code readability and consistency.  

image

### DIFF
--- a/src/batterymonitor.h
+++ b/src/batterymonitor.h
@@ -45,6 +45,9 @@
     #define ADCVoltageMax 1.0
 #endif
 
+#ifndef BATTERY_SHIELD_RESISTANCE
+    #define BATTERY_SHIELD_RESISTANCE 180.0
+#endif
 #ifndef BATTERY_SHIELD_R1
     #define BATTERY_SHIELD_R1 100.0
 #endif
@@ -56,7 +59,7 @@
     #ifndef PIN_BATTERY_LEVEL
         #error Internal ADC enabled without pin! Please select a pin.
     #endif
-    // Wemos D1 Mini has an internal Voltage Divider with R1=220K and R2=100K > this means, 3.3V analogRead input voltage results in 1023.0
+    // Wemos D1 Mini has an internal Voltage Divider with R1=100K and R2=220K > this means, 3.3V analogRead input voltage results in 1023.0
     // Wemos D1 Mini with Wemos Battery Shield v1.2.0 or higher: Battery Shield with J2 closed, has an additional 130K resistor. So the resulting Voltage Divider is R1=220K+100K=320K and R2=100K > this means, 4.5V analogRead input voltage results in 1023.0
     // ESP32 Boards may have not the internal Voltage Divider. Also ESP32 has a 12bit ADC (0..4095). So R1 and R2 can be changed.
     // Diagramm: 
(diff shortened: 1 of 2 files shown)


Eirenliel:
also i'm thinking here... production slimes have different ADC resistors and the firmware is the same for them... um...

production-adc

Vyolex:
Technically it doesn't matter if the ratio are the same. The common configuration on DIY would equate to 400 and 100 so the 40.2 and 10 should be close enough. But that might be something we should fix, even if only for readability. I'm fairly certain I mentioned in my last Pr that these really should be in the defines or definitions file for that exact reason :p

src/batterymonitor.h
```diff
@@ -62,7 +62,7 @@
     // Diagramm: 
     //   (Battery)--- [BATTERY_SHIELD_RESISTANCE] ---(INPUT_BOARD)---  [BATTERY_SHIELD_R2] ---(ESP_INPUT)--- [BATTERY_SHIELD_R1] --- (GND)
```
unlogisch04, line 63: Your diagram is different than this one here.
I don't think we should change anything.
There is not clear definition for R1 and R2 how it have to be.

Eirenliel:
> Technically it doesn't matter if the ratio are the same. The default defines as they stand would equate to 400 and 100 so the 40.2 and 10 should be close enough. But that might be something we should fix, even if only for readability. I'm fairly certain I mentioned in my last Pr that these really should be in the defines or definitions file for that exact reason :p

but they aren't the same we don't change firmware for battery sense before uploading 😭

Vyolex:
> but they aren't the same we don't change firmware for battery sense before uploading 😭

I know we don't change the values, but the ratio is what is important;

(40.2+10)/10 = 5.02
(400 + 100)/100 = 5

The difference is extremely minimal (only 0.4%), idk what deviation is allowed on the resistors used within the official trackers but this is much less than the recommended 1% for DIY use. 

That being said I fully agree if you would change it for the future slimes, I'm just trying to stress that it is not a very big issue.

src/batterymonitor.h
Vyolex: > Wemos D1 Mini has an internal Voltage Divider with R1=220K and R2=100K > this means, 3.3V analogRead input voltage results in 1023.0

Then this line has conflicting information, I agree that it is just naming and doesn't really matter. I just wanted to make it more consistent with documentation that is available. I am completely fine reverting all changes and just adapting the values in this single line above though.

Eirenliel:
oh right i forgor that it adds +180, so we gucci.

Vyolex:
Reverted earlier changes but adapted comments so everything is consistent.

Added board specific resistor values with override option (and fallback).

Eirenliel (approved):

Eirenliel:
What I'm not sure about is that NODEMCU actually doesn't have battery shield.

PR #228, opened 2023-02-06
feat: commit hash

unlogisch04:
## Change

This pull request adds the Git commit hash in the built firmware.

## Notes:

The web flasher needs a new implementation, this should not be merged without the change.

⚠️ Needs testing ⚠️


Eirenliel (approved):

ButterscotchV:
Added this branch to my site (https://slimevr-firmware.bscotch.ca/) as [unlogisch04/feat_commitid](https://github.com/unlogisch04/SlimeVR-Tracker-ESP/tree/feat_commitid) to be used for testing.

I can confirm this breaks on the web firmware tool. Check the attached log for more info:
[firmware-tool-log.log](https://github.com/SlimeVR/SlimeVR-Tracker-ESP/files/10702421/firmware-tool-log.log)

ButterscotchV (changes requested):
This shouldn't crash if not in a git repo (ex. downloading a release zip, etc), even though it's not much of an expected case, it's good to keep it working anyways. Maybe a big warning instead?

For firmware tool compatability, support should be added for reading the commit ID from an environment variable, as we have the metadata but no way to pass it in.

unlogisch04:
Oh i did forgot, that it can be out of a git repo.
Thanks for testing.

TheDevMinerTV:
@ButterscotchV could you check again? I've added error handling to the script

ButterscotchV:
Tested again, doesn't compile
```py
File "scripts/get_git_commit.py", line 22
    print(f"'-DGIT_REV=\"{revision}\"'")
                                      ^
SyntaxError: invalid syntax
```
[firmware-tool-log.log](https://github.com/SlimeVR/SlimeVR-Tracker-ESP/files/11683240/firmware-tool-log.log)

nekomona:
> Tested again, doesn't compile
> 
> ```python
> File "scripts/get_git_commit.py", line 22
>     print(f"'-DGIT_REV=\"{revision}\"'")
>                                       ^
> SyntaxError: invalid syntax
> ```
> 
> [firmware-tool-log.log](https://github.com/SlimeVR/SlimeVR-Tracker-ESP/files/11683240/firmware-tool-log.log)

That could be from an old version of python, f-strings are supported from python 3.6. Working fine with ubuntu and PlatformIO installed last month

nekomona (changes requested):
Tested with these two patches with and without git repo

scripts/get_git_commit.py
```diff
@@ -1,14 +1,22 @@
...
         .strip()
         .decode("utf-8")
     )
-except:
-    pass
+except Exception:
+    print("==========================================================================")
+    print("  WARNING: You are not in a Git repository or do not have Git installed.")
+    print("           The version string will be incorrect.")
+    print("==========================================================================")
```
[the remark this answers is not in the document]
TheDevMinerTV: Ohhh, you're right, thank you :/

scripts/get_git_commit.py
```diff
@@ -14,9 +14,6 @@
         .decode("utf-8")
     )
 except Exception:
-    print("==========================================================================")
-    print("  WARNING: You are not in a Git repository or do not have Git installed.")
-    print("           The version string will be incorrect.")
-    print("==========================================================================")
+    revision = "NOT_GIT"
```
ButterscotchV, line 17: Won't this remove any git revision set through an environment variable?

nekomona: Not sure about if it's about customized case where there's an extra line in the `platformio.ini`. At that case since `platformio.ini` is to be modified anyway, could just comment out the line calling this Python script and write the line for `GIT_REV`.

For other cases the script just set `GIT_REV` string to `"NOT_GIT"` when can't get current commit hash.

ButterscotchV: The web firmware tool has the git commit hash, but it doesn't use it as a git repo, so this will always throw an exception. It's best that the web firmware tool doesn't modify any files it doesn't need to as that's more to maintain, but setting an environment variable is really easy.

scripts/get_git_commit.py
```diff
@@ -0,0 +1,19 @@
...
+    revision = env_rev
+
+try:
+    revision = (
+        subprocess.check_output(["git", "rev-parse", "HEAD"])
+        .strip()
+        .decode("utf-8")
+    )
+except Exception:
+    revision = "NOT_GIT"
```
nekomona, line 17:
```suggestion
revision = ""

env_rev = os.environ.get("GIT_REV")
if not env_rev is None and env_rev != "":
    revision = env_rev
else:
	try:
		revision = (
			subprocess.check_output(["git", "rev-parse", "HEAD"])
			.strip()
			.decode("utf-8")
		)
	except Exception:
		revision = "NOT_GIT"
```

ButterscotchV:
Oh also I've tested and it does appear to compile. I've added support for the `GIT_REV` environment variable on my web firmware tool fork, so that can also be tested hopefully (I hope I did it right -w-).

scripts/get_git_commit.py
TheDevMinerTV: Mixed tabs and spaces

nekomona: oops I've just typed within the browser ∑0w0. It's fixed now

scripts/get_git_commit.py
TheDevMinerTV: fixed with the latest commit

ButterscotchV (approved):
Works on web firmware tool perfectly 👍

TheDevMinerTV (approved):
LGTM

@Eirenliel please merge

Eirenliel (approved):

PR #233, opened 2023-03-14
Fix AUX trackers inconsistent tracker ID issue.

ThreadOfFate:
This fixes a issue that would be seen as the server swapping the main and aux imu.
It bug will also swap the calibration between main and aux imus resulting in worse tracking.

### DIFF
--- a/src/sensors/SensorManager.cpp
+++ b/src/sensors/SensorManager.cpp
@@ -42,10 +42,6 @@ namespace SlimeVR
             uint8_t firstIMUAddress = 0;
             uint8_t secondIMUAddress = 0;
 
-            //TODO check sensorID the 0 in (ErroneousSensor(0, IMU);) to sync with the address
-            //what is important is that is consitent, so having the sensorid 0 tied to all the default addresses will fix it
-            //create defines for all the address values set them to the address and then check if the
-
             {
                 firstIMUAddress = I2CSCAN::pickDevice(PRIMARY_IMU_ADDRESS, SECONDARY_IMU_ADDRESS, true);
                 uint8_t sensorID = 0;
--- a/src/sensors/SensorManager.cpp
+++ b/src/sensors/SensorManager.cpp
@@ -31,6 +31,7 @@
 #include "bmi160sensor.h"
 #include "icm20948sensor.h"
 #include "ErroneousSensor.h"
+#include "sensoraddresses.h"
 
 namespace SlimeVR
 {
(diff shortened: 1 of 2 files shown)


src/sensors/sensoraddresses.h
```diff
@@ -0,0 +1,21 @@
...
+    #if IMU == IMU_BNO080 || IMU == IMU_BNO085 || IMU == IMU_BNO086
+        #define PRIMARY_IMU_ADDRESS 0x4A
+        #define SECONDARY_IMU_ADDRESS 0x4B
+    #elif IMU == IMU_BNO055
+        #define PRIMARY_IMU_ADDRESS 0x29
+        #define SECONDARY_IMU_ADDRESS 0x28
+    #elif IMU == IMU_MPU9250 || IMU == IMU_BMI160 || IMU == IMU_MPU6500 || IMU == IMU_MPU6050 || IMU == IMU_ICM20948
+        #define PRIMARY_IMU_ADDRESS 0x68
+        #define SECONDARY_IMU_ADDRESS 0x69
+    #endif
```
0forks, line 16: This will be breaking, people mix BNO with a different IMU type, it needs to account for `SECOND_IMU`.

Eirenliel: Agree

ThreadOfFate: Pushed a commit there that should cover mixed IMU types, but I will need someone else to test it.

Eirenliel (approved):

src/sensors/SensorManager.cpp
```diff
@@ -41,78 +42,73 @@ namespace SlimeVR
...
-#elif IMU == IMU_BNO055
-                firstIMUAddress = I2CSCAN::pickDevice(0x29, 0x28, true);
-#elif IMU == IMU_MPU9250 || IMU == IMU_BMI160 || IMU == IMU_MPU6500 || IMU == IMU_MPU6050 || IMU == IMU_ICM20948
-                firstIMUAddress = I2CSCAN::pickDevice(0x68, 0x69, true);
-#else
-#error Unsupported primary IMU
-#endif
+                firstIMUAddress = I2CSCAN::pickDevice(PRIMARY_IMU_ADDRESS_ONE, PRIMARY_IMU_ADDRESS_TWO, true);
+                uint8_t sensorID = 0;
+                if(sharedIMUAddresses && firstIMUAddress != PRIMARY_IMU_ADDRESS_ONE)
```
ButterscotchV, line 49: If `firstIMUAddress` is `0`, then it will match this condition, meaning that the `sensorID` will be changed to `1` for an erroneous sensor and could cause confusion about which device is causing the error.

src/sensors/SensorManager.cpp
```diff
@@ -41,78 +42,73 @@ namespace SlimeVR
...
-#elif SECOND_IMU == IMU_BNO055
-                secondIMUAddress = I2CSCAN::pickDevice(0x28, 0x29, false);
-#elif SECOND_IMU == IMU_MPU9250 || SECOND_IMU == IMU_BMI160 || SECOND_IMU == IMU_MPU6500 || SECOND_IMU == IMU_MPU6050 || SECOND_IMU == IMU_ICM20948
-                secondIMUAddress = I2CSCAN::pickDevice(0x69, 0x68, false);
-#else
-#error Unsupported secondary IMU
-#endif
+                secondIMUAddress = I2CSCAN::pickDevice(SECONDARY_IMU_ADDRESS_TWO, SECONDARY_IMU_ADDRESS_ONE, false);
+                uint8_t sensorID = 1;
+                if(sharedIMUAddresses && secondIMUAddress != SECONDARY_IMU_ADDRESS_TWO)
```
ButterscotchV, line 83: Same as the last comment, if `secondIMUAddress` is `0`, then it will match this condition, meaning that the `sensorID` will be changed to `0` for an erroneous sensor and could cause confusion about which device is causing the error.

ButterscotchV (changes requested):

src/sensors/SensorManager.cpp
ThreadOfFate: In pratice I don't see this being an issue.
firstIMUAddress is 0 when the ic2 scan fails to find any imus.
If the addresses are shared then it means no imus are detected at all.
If the addresses are not shared then sensorID is never set to 1.

ButterscotchV: Thanks for explaining, looks fine then.

ButterscotchV (approved):

0forks:
Maybe needs a PSA that recalibration will be needed for people who didn't wire addresses according to the docs (9250 or BMI), as it'll swap their sensor ids, even when only 1 IMU is present.

ThreadOfFate:
> Maybe needs a PSA that recalibration will be needed for people who didn't wire addresses according to the docs (9250 or BMI), as it'll swap their sensor ids, even when only 1 IMU is present.

PSA sounds like a good idea.
We should cover
-that tracker assignment may switch in the server
-calibration may have swapped between sensors, before or because of this update and that it is recommend in general to recalibrate

PR #236, opened 2023-03-30
Replace BMI160 Library with fork of DFRobot

RexaOnGit:
I have made it as far as waiting for server connection in testing this refactor. There are likely more bugs with rhe new library, but I can't test them without server connection.

I'm not sure why server connection doesn't work, since I didn't change any network code. I think at this point I need help debugging and reviewing.

I expect several commits to be necessary before this refactor is ready for merge.


0forks:
I only don't like some inconsistencies that this brings to the FIFO defines and minor clarity regressions in naming.

Haven't tried yet, will test later.

RexaOnGit:
> I only don't like some inconsistencies that this brings to the FIFO defines and minor clarity regressions in naming. And the fact that this makes single readouts fatter - previously if it requested only gyro it read only gyro, not gyro+accel, same with accel.

This library was clearly setup with different useage in mind - there are a lot of features we don't need and a lack of many features we do need. The mess of private/public overloads is also quite confusing.

I didn't want to majorly trim down or refactor the library until I got to a working state anyway.

RexaOnGit:
There were too many problems with the DF robot library. I am abandoning and closing this pull request. Will be back with a different one later - building a new BMI160 library from scratch this time.

PR #242, opened 2023-05-13
Add ICM-42688 imu and MMC5983MA magnetometer

sctanf:


sctanf:
both ICM-42688 and MMC5983MA are working, data can be received by the tracker, and it looks correct. Not able to test calibration or fusion of the data, the tracker completely died (I shorted something?) before i could do more and I give up.

Vyolex (approved):
Can't test IMU functionality but the code itself looks ok

Eirenliel:
This needs to be updated

Eirenliel (approved):

PR #246, opened 2023-05-24
Fix 9250 loop

castlemccloud:
Previous refactoring of the code rolled both the assignment of the temp variable and the calibration matrix multiplication into a single loop.

This causes issue because every iteration of the loop for the matrix multiplication uses every value of temp, and it is not completely initialized until the final loop iteration.

The solution is to split the loop into two.

### DIFF
--- a/src/sensors/mpu9250sensor.cpp
+++ b/src/sensors/mpu9250sensor.cpp
@@ -398,8 +398,9 @@ void MPU9250Sensor::parseMagData(int16_t data[3]) {
     float temp[3];
 
     //apply offsets and scale factors from Magneto
-    for (unsigned i = 0; i < 3; i++)
+    for (unsigned i = 0; i < 3; i++) {
         temp[i] = (Mxyz[i] - m_Calibration.M_B[i]);
+    }
     
     for (unsigned i = 0; i < 3; i++) {
         #if useFullCalibrationMatrix == true
@@ -419,8 +420,9 @@ void MPU9250Sensor::parseAccelData(int16_t data[3]) {
     float temp[3];
 
     //apply offsets (bias) and scale factors from Magneto
-    for (unsigned i = 0; i < 3; i++)
-        temp[i] = (Axyz[i] - m_Calibration.A_B[i]);
+    for (unsigned i = 0; i < 3; i++) {
+        temp[i] = (Mxyz[i] - m_Calibration.M_B[i]);
+    }
     
     for (unsigned i = 0; i < 3; i++) {
         #if useFullCalibrationMatrix == true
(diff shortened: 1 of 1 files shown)


kitlith:
Yep, this is:
 - Unambiguously a bug (that is still present on main)
 - Definitely my fault

I'll rebase it and apply a style tweak real quick.

TheDevMinerTV (approved):

src/sensors/mpu9250sensor.cpp
```diff
@@ -398,8 +398,10 @@ void MPU9250Sensor::parseMagData(int16_t data[3]) {
     float temp[3];
 
     //apply offsets and scale factors from Magneto
-    for (unsigned i = 0; i < 3; i++) {
+    for (unsigned i = 0; i < 3; i++)
         temp[i] = (Mxyz[i] - m_Calibration.M_B[i]);
```
kitlith, line 402:
```suggestion
    for (unsigned i = 0; i < 3; i++) {
        temp[i] = (Mxyz[i] - m_Calibration.M_B[i]);
    }
```

src/sensors/mpu9250sensor.cpp
```diff
@@ -417,8 +419,10 @@ void MPU9250Sensor::parseAccelData(int16_t data[3]) {
     float temp[3];
 
     //apply offsets (bias) and scale factors from Magneto
-    for (unsigned i = 0; i < 3; i++) {
+    for (unsigned i = 0; i < 3; i++)
         temp[i] = (Axyz[i] - m_Calibration.A_B[i]);
```
kitlith, line 423:
```suggestion
    for (unsigned i = 0; i < 3; i++) {
        temp[i] = (Mxyz[i] - m_Calibration.M_B[i]);
    }
```

kitlith (approved):
Overall, this is a straightforward fix, and more or less what was done before my incorrect refactor.

I'm adding braces around the for loop -- really it's just a stylistic thing, my excuse is that it's less prone to human error.

(Technically there are still some differences, but no functional differences. Wouldn't be surprised if the optimizer found them to be identical.)

src/sensors/mpu9250sensor.cpp
```diff
@@ -419,8 +420,9 @@ void MPU9250Sensor::parseAccelData(int16_t data[3]) {
     float temp[3];
 
     //apply offsets (bias) and scale factors from Magneto
-    for (unsigned i = 0; i < 3; i++)
-        temp[i] = (Axyz[i] - m_Calibration.A_B[i]);
+    for (unsigned i = 0; i < 3; i++) {
+        temp[i] = (Mxyz[i] - m_Calibration.M_B[i]);
```
kitlith, line 424:
```suggestion
        temp[i] = (Axyz[i] - m_Calibration.A_B[i]);
```

TheDevMinerTV (approved):

Nuzair46 (approved):

Eirenliel (approved):

PR #248, opened 2023-06-03
Unifying Sensor Fusion Code to Abstract Between Sensors And Fusion Algorithms

nekomona:
The codes for 6/9-axis fusion in BMI160/MPU9250, the codes for linear acceleration extraction in BMIs/MPUs/ICMs, and the codes for DMP+MAG fusion in MPU9250 are all abstracted out and unified into the new `SensorFusion` classes. 

After this modification, sensors with raw gyro/accel/mag data could use the `SensorFusion` base class to do all the fusion, including upcoming sensors. If some of these sensors also need rest detection to do dynamic calibration like the BMI160, they could use the `SensorFusionRestDetect`, which could take detection result from vqf or add additional detector to achieve that. MPUs using DMP have a separate class `SensorFusionDMP`, which abstracts the dmpmag algorithm and provides a similar interface to the raw `SensorFusion` class.

With the abstraction, it could be more straightforward to introduce new algorithms or new sensors. Besides, it enables the combination previously not written like *MPU6050 + QMC5883L + VQF*, and future sensors will have access to all algorithms automatically.

I have tested it with *MPU6050+QMC5883L* and *BMI160+QMC5883L* in 6/9-axis modes. ICM20948 is not tested but should do exactly the same as before, and no modification is made for BNOs as they do no fusions on the MCU.


Eirenliel (approved):
Woah looks great, we need this released as beta for the community to test before merging

src/sensors/SensorFusionRestDetect.h
```diff
@@ -0,0 +1,55 @@
...
+                void updateGyro(sensor_real_t Gxyz[3], sensor_real_t deltat);
+            #endif
+        protected:
+            #if !SENSOR_WITH_REST_DETECT
+                SensorRestDetectionParams restDetectionParams {};
+                RestDetection restDetection;
+            #endif
+            
+		};
+	}
```
TheDevMinerTV, line 52: Please replace these tabs with spaces

TheDevMinerTV (changes requested):
Please add the missing ending newlines and fix those tabs to be spaces. Otherwise looks fine.

src/sensors/SensorFusionRestDetect.h
nekomona: All tabs in new code are replaced, ending newlines are checked and fixed in the latest commit -w-/

lib/magneto/madgwick.hpp
```diff
@@ -165,4 +166,4 @@ void madgwickQuaternionUpdate(float q[4], float ax, float ay, float az, float gx
 	q[1] *= recipNorm;
 	q[2] *= recipNorm;
 	q[3] *= recipNorm;
```
TheDevMinerTV: there's still quite a lot of tabs here

src/sensors/SensorFusionRestDetect.h
```diff
@@ -0,0 +1,55 @@
...
+        #endif
+
+        class SensorFusionRestDetect : public SensorFusion
+        {
+        public:
+            SensorFusionRestDetect(float gyrTs, float accTs=-1.0, float magTs=-1.0)
+                : SensorFusion(gyrTs, accTs, magTs)
+            #if !SENSOR_WITH_REST_DETECT
+                , restDetection(restDetectionParams, gyrTs,
+                                ((accTs<0) ? gyrTs : accTs) )
```
TheDevMinerTV, line 35:
```suggestion
                                ((accTs<0) ? gyrTs : accTs))
```

src/sensors/SensorFusionRestDetect.cpp
```diff
@@ -0,0 +1,34 @@
+#include "SensorFusionRestDetect.h"
+
+namespace SlimeVR
+{
+    namespace Sensors
+    {
+        void SensorFusionRestDetect::updateAcc(sensor_real_t Axyz[3], sensor_real_t deltat)
+        {
```
TheDevMinerTV, line 8:
```suggestion
        {
```

TheDevMinerTV (changes requested):
Just some small style fixes needed, looks good otherwise

src/sensors/SensorFusionRestDetect.h
nekomona: I might prefer to leave the space here, so that the double right bracelets won't be confused to be paring with the double left bracelets before the ? operation

TheDevMinerTV: I guess or do it like this,
```suggestion
                                (accTs<0) ? gyrTs : accTs)
```

nekomona: That's less confusing as the first bracelet is clear to be in pairs so guess I'll take that. I've been used to guard every ? operator in its own bracelets and leave spaces in complex lines, but that one looks fairly simple.

TheDevMinerTV (approved):
LGTM, just waiting for beta testers.

src/sensors/SensorFusion.h
```diff
@@ -0,0 +1,122 @@
...
+            bool magExist = false;
+            sensor_real_t qwxyz[4]{1.0f, 0.0f, 0.0f, 0.0f};
+            bool updated = false;
+
+            bool gravityReady = false;
+            sensor_real_t vecGravity[3]{0.0f, 0.0f, 0.0f};
+            bool linaccelReady = false;
+            sensor_real_t linAccel[3]{0.0f, 0.0f, 0.0f};
+            #if ESP32
+            sensor_real_t linAccel_guard;   // Temporary patch for some weird ESP32 bug
```
TheDevMinerTV, line 116: What exactly does this do?

nekomona: It just adds another spare memory space to avoid writings exceeding the end of `linAccel` spill out to other sensitive data like memory allocation magics or so. I've checked the code and disassembly to confirm that there is no code writing exceed `linAccel[2]`, but the bug does exist in ESP32, and the weird patch does fix it. Maybe that'll be something hidden in the libraries but just can't find them.

TheDevMinerTV: huh okay

TheDevMinerTV (approved):

kitlith (commented):
The architecture here feels weird in general, and not what I would have expected, but getting something with the exact semantics I'm looking for seems tricky, or requires a c++20 feature that I'm not very familiar with (concepts), so we do not need to change it right now.

src/sensors/SensorFusion.cpp
```diff
@@ -0,0 +1,173 @@
...
+            updateAcc(Axyz, deltat);
+            updateGyro(Gxyz, deltat);
+        }
+
+        void SensorFusion::update9D(sensor_real_t Axyz[3], sensor_real_t Gxyz[3], sensor_real_t Mxyz[3], sensor_real_t deltat)
+        {
+            updateMag(Mxyz, deltat);
+            updateAcc(Axyz, deltat);
+            updateGyro(Gxyz, deltat);
+        }
```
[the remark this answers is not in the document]
nekomona: I was trying to recreate the behavior in [bmi160sensor.cpp](https://github.com/SlimeVR/SlimeVR-Tracker-ESP/blob/d970e7cbc4f648febcc4bb77a42f931939521adb/src/sensors/bmi160sensor.cpp#L491-L514) here, recreating that sequence with each gyro sample raise a new quaternion.

I'm not very sure about what would happen if I reverse that sequence so I tried to keep all the behaviors unchanged, including this one. XP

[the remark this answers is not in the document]
nekomona: btw what's the architecture you're looking for? I could check if some of them could be migrated.

PR #249, opened 2023-06-03
Add Per-Sensor Descriptor and Support Multiple IMUs

nekomona:
This modification introduces sensor descriptor to specify sensor config in a clearer way. Currently it contains sensor types, address, pins and extra configs like axis remapping.  It may extend to more parameters like magnetometer type in the future.

Descriptors are written in macros in the following format thanks to @yoyobuae :
```c++
#define IMU_DESC_LIST \
             IMU_DESC_ENTRY(IMU_MPU6050, 0x68, DEG_270, PIN_IMU_SCL, PIN_IMU_SDA) \
             IMU_DESC_ENTRY(IMU_BMI160,  0x69, DEG_270, PIN_IMU_SCL, PIN_IMU_SDA, AXIS_REMAP_DEFAULT)
```
They will be expanded to sensor building and detecting in the `SensorManager`. Sensor ID is assigned by the sequence of sensors in the list, and other parameters will be passed to building functions. This creates a direct connection between sensor id, sensor address, and sensor pins / configs in the same place.

The new descriptor also allows to add more than 2 sensors by appending more lines to the descriptor. To implement this, each sensor config contains the pin for that sensor. Internal pinmux inside ESPs are used to switch between different sensors, so more than 2 sensors with the same base address could be hooked up to different pins without extra hardware.

The code is highly experimental but attempted to be compatible with previous configs. When not specified, the descriptors will be generated from previous defines with matching address and ids. If a previous config uses two types of sensors but with incorrect address pin connection, the new code won't try the other address to avoid mismatch between config and assigned sensor id.  

Axis remapping for BMI160 may not be fully compatible with previous configs. Now axis mapping is described by a number generated from defines. The sensor will use the number for no remapping by default to match the previous default behavior. Each BMI160 could have its own number for a different remapping in its descriptor. To migrate for this patch, previous changes in `bmi160_defines.h` for axis mapping will need to be rewritten in the new descriptors unless no remapping is used. 

I have tested this in 2-sensor setups with combination of BMI160 and MPU6050s. An experimental setup driving 8 MPU6050s are also tested and works with only additional lines in the descriptor. INT pin for BNOs may need to be tested but should be working as the only difference is in passing pin numbers.


src/sensors/SensorManager.h
```diff
@@ -37,26 +37,37 @@ namespace SlimeVR
...
 
             void setup();
             void postSetup();
 
             void update();
-
-            Sensor *getFirst() { return m_Sensor1; };
-            Sensor *getSecond() { return m_Sensor2; };
+            
+            Sensor  ** const getSensors() { return m_Sensors; };
```
TheDevMinerTV, line 59: Not sure if this should be a std::vector or something, I feel like this could be worrysome with memory safety. (I don't know C++ that well though)

src/sensors/SensorManager.cpp
```diff
@@ -33,109 +33,171 @@
...
-            uint8_t secondIMUAddress = 0;
+            m_Logger.trace("Building IMU with: id=%d,\n\
+                            imuType=0x%02X, address=0x%02X, rotation=%f,\n\
+                            sclPin=%d, sdaPin=%d, extraParam=%d",
+                            sensorID,
+                            imuType, address, rotation,
+                            sclPin, sdaPin, extraParam);
+
+            // Now start detecting and building the IMU
+            Sensor* sensor = NULL;
```
TheDevMinerTV, line 54:
```suggestion
            Sensor* sensor = nullptr;
```

src/sensors/SensorManager.cpp
```diff
@@ -33,109 +33,171 @@
...
+            
+            // Clear and reset I2C bus for each sensor upon startup
+            I2CSCAN::clearBus(sdaPin, sclPin);
+            swapI2C(sclPin, sdaPin);
+
+            if (I2CSCAN::isI2CExist(address)) {
+                m_Logger.trace("IMU %d found at address 0x%02X", sensorID, address);
+            } else {
+                sensor = new ErroneousSensor(sensorID, imuType);
```
TheDevMinerTV, line 63: Couldn't we just return the sensor here?

src/sensors/SensorManager.cpp
```diff
@@ -33,109 +33,171 @@
...
+                sensorID++;                                          \
+                if (sensor->isWorking()) {                           \
+                    m_Logger.info("Sensor %d configured", sensorID); \
+                    activeSensorCount++;                             \
+                }                                                    \
+            }
+            // Apply descriptor list and expand to entrys
+            IMU_DESC_LIST;
+
+#undef IMU_DESC_ENTRY
```
TheDevMinerTV, line 159: This feels a bit hacky, but I guess it works

TheDevMinerTV (approved):
LGTM, I've still added some comments though

src/sensors/SensorManager.h
nekomona: I'm not sure if it's good to use dynamic allocated memory here, but that should do the job (and fix many other weird-looking loops)

src/sensors/SensorManager.cpp
nekomona: That is for reproducing the previous behavior in L54-L57 here:
https://github.com/SlimeVR/SlimeVR-Tracker-ESP/blob/48d87a327eb8bc1bd4d4174a3aa549c18011937f/src/sensors/SensorManager.cpp#L54-L57

src/sensors/SensorManager.cpp
nekomona: Yes it's kind of preprocessor spells here. That's also my first time seeing it used in an MCU project.

src/sensors/SensorManager.h
TheDevMinerTV:
```suggestion
            Sensor ** const getSensors() { return m_Sensors; };
```
oh yeah, there was a double space in here

nekomona: That array is replaced with vector in the latest commit and thus the double space is gone -w-/

src/sensors/SensorManager.cpp
```diff
@@ -33,109 +33,171 @@
...
-                    m_Sensor2 = new MPU9250Sensor(sensorID, secondIMUAddress, SECOND_IMU_ROTATION);
-#elif SECOND_IMU == IMU_BMI160
-                    m_Sensor2 = new BMI160Sensor(sensorID, secondIMUAddress, SECOND_IMU_ROTATION);
-#elif SECOND_IMU == IMU_MPU6500 || SECOND_IMU == IMU_MPU6050
-                    m_Sensor2 = new MPU6050Sensor(sensorID, SECOND_IMU, secondIMUAddress, SECOND_IMU_ROTATION);
-#elif SECOND_IMU == IMU_ICM20948
-                    m_Sensor2 = new ICM20948Sensor(sensorID, secondIMUAddress, SECOND_IMU_ROTATION);
-#endif
-                }
+        void SensorManager::swapI2C(uint8_t sclPin, uint8_t sdaPin)
```
Eirenliel, line 109:
```suggestion
        // TODO Make it more generic in the future and move another place (abstract sensor interface)
        void SensorManager::swapI2C(uint8_t sclPin, uint8_t sdaPin)
```

src/sensors/SensorManager.h
```diff
@@ -56,12 +57,12 @@ namespace SlimeVR
 
             void update();
             
-            Sensor  ** const getSensors() { return m_Sensors; };
+            std::vector<Sensor *> & getSensors() { return m_Sensors; };
```
TheDevMinerTV, line 60: We should probably use smart pointers for this, but again, I'm no C++ expert
Here's some MS docs about them: https://learn.microsoft.com/en-us/cpp/cpp/smart-pointers-modern-cpp?view=msvc-170#c-standard-library-smart-pointers

src/sensors/SensorManager.cpp
TheDevMinerTV: Oh, I just realized, you're never calling `motionSetup()` on this sensor.

Eirenliel (approved):
Another banger. Really want to merge this and the other one, and also merge my refactoring so there are less conflicts lol

src/serial/serialcommands.cpp
```diff
@@ -145,16 +139,16 @@ namespace SerialCommands {
...
-                sensor1->hadData ? "true" : "false"
+                "[TEST] Sensor[0]: %s (%.3f %.3f %.3f %.3f) is working: %s, had data: %s",
+                getIMUNameByType(sensor0->getSensorType()),
+                UNPACK_QUATERNION(sensor0->getQuaternion()),
+                sensor0->isWorking() ? "true" : "false",
+                sensor0->hadData ? "true" : "false"
             );
-            if(!sensor1->hadData) {
+            if(!sensor0->hadData) {
                 logger.error("[TEST] Sensor 1 didn't send any data yet!");
```
Eirenliel, line 152:
```suggestion
                logger.error("[TEST] Sensor 0 didn't send any data yet!");
```

src/serial/serialcommands.cpp
```diff
@@ -145,16 +139,16 @@ namespace SerialCommands {
...
             } else {
                 logger.info("[TEST] Sensor 1 sent some data, looks working.");
```
Eirenliel, line 154:
```suggestion
                logger.info("[TEST] Sensor 0 sent some data, looks working.");
```

Eirenliel (approved):

Eirenliel:
Firmware webflasher would need to be rebuilt for this... @ButterscotchV @loucass003...

src/sensors/SensorManager.cpp
```diff
@@ -114,8 +140,41 @@ namespace SlimeVR
...
+                    else {
+                        // Reset HWI2C to avoid being affected by I2CBUS reset
+                        Wire.end();
+                    }
+                    // Disconnect pins from HWI2C
+                    pinMode(activeSCL, INPUT);
+                    pinMode(activeSDA, INPUT);
+
+                    if (running) {
+                        i2c_set_pin(I2C_NUM_0, sdaPin, sclPin, false, false, I2C_MODE_MASTER);
```
unlogisch04, line 125: Is there a reason to deactivate the pull ups for SDA/SCL?

PR #253, opened 2023-06-23
add new battery mappings

sctanf:
measure an actual existing battery and use those values for mapping instead of some random internet graph
maybe this is better?


sctanf:
> Hi I'm missing documentation. In what case do i have to set what option, for what reason? REG_BUCK REG_LDO REG_LEGACY

reg -> regulator, basically what did the board have for power regulation
official slimevr has a buck converter circuit, basically all diy slimes have a linear ldo regulator (like d1 mini)
setting buck or ldo uses the appropriate voltage mapping for either regulator type, and legacy is just the old behavior

unlogisch04:
Ok, thanks. 
This should go into the Code too. Imho
Also how did you get this Values?
Im asking, because to me this are numbers i have no clue how to reproduce.

Im not sure if that would need to go into the Code, but if someone needs to make a change, the numbers could be reproduce. (For different circuit)

src/batterymonitor.h
```diff
@@ -54,6 +54,10 @@
     #define BATTERY_SHIELD_R2 220.0
 #endif
 
+#ifndef BATTERY_REGULATOR
+    #define BATTERY_REGULATOR REG_LDO
```
Vyolex, line 58: To keep backwards compatibility the fallback should be legacy behavior imo.
```suggestion
    #define BATTERY_REGULATOR REG_LEGACY
```

src/batterymonitor.cpp
Vyolex, line 1: Will verify the curves with a test, in theory they look correct.

Vyolex (changes requested):
Code looks good to me, just some minor nitpicks.

src/batterymonitor.h
Vyolex: I mean fair, but then why even keep it around in the first place. At the very least it should be official slime option as default then imo. I'll be setting up a test today.

sctanf: official slime uses a buck regulator, which will draw roughly constant power, the discharge characteristic is different than most diy boards which have a linear ldo regulator, which will draw roughly constant current

src/batterymonitor.cpp
Vyolex: looks like i messed up the test a bit so charging the slime and will try again tomorrow

sctanf: Did you ever get a chance to test this?

ButterscotchV (commented):
Looks fine, I have absolutely no way to test this, so I don't have much useful input besides that it should work. Why no `REG_BUCK` for legacy SlimeVR design? I don't really know anything about that board so I'm not sure if it's needed.

src/batterymonitor.cpp
Vyolex: Didn't get to doing this, had some stuff come up when I was doing it, never finished and forgot about it.

sctanf:
> Looks fine, I have absolutely no way to test this, so I don't have much useful input besides that it should work. Why no `REG_BUCK` for legacy SlimeVR design? I don't really know anything about that board so I'm not sure if it's needed.

It would be helpful if someone was able to test i think

I'm not sure what the regulator is on legacy/dev board but it probably isn't too important, if it has a buck regulator then it can be added then

Vyolex:
> It would be helpful if someone was able to test i think
   The curve obtained should be better but I could make no tests with specificity esp hardware

I have some time again and I can run a test soon to see if the curve looks good. Wanted to test something else along with this either way. @sctanf could you update the branch and then I'll run some tests this weekend orso on the official hardware.

Vyolex:
Ok thank you, will test next weekend 😃

Vyolex:
Wanted to run a test, but this firmware currently doesn't work and crashes the slime. Confirmed working on main so I'm guessing the merge is to blame.

Vyolex:
@ImUrX @sctanf is this still something we want to work on? Now that Uriel got the proper measurements.

sctanf:
it should be updated with new measurements for official trackers, then keep legacy mapping for now because new one was not tested for diy

unlogisch04:
I think we could map it to the Board. So if the board is SlimeVR then use the mapping from SlimeVR and else the legacy one?

Eirenliel:
This is too stale now. I think if we really want to improve this, it should be per board as suggested, and it should be in board defines somehow...

PR #255, opened 2023-06-30
Add Remote Command Over WiFi

nekomona:
This PR add a remote command server running in parallel with the serial command, sharing the same command buffer, command lists, and logger outputs. All the data printed through `Logger` will be forked to the remote when connected, while the ones using `Serial` won't be forked.

The remote command server could be accessed through telnet clients. Would be easy to integrate into server since it use a simple TCP socket to transmit the same data as serial commands do.

Currently this remote command mostly benefits for checking TCAL status remotely, as other commands are mostly for debugging instead of functions. This could be used for future function commands which are too minor to occupy a packet type or doesn't fit well in a UDP packet like largely varying data length.

### DIFF
--- a/src/serial/serialcommands.cpp
+++ b/src/serial/serialcommands.cpp
@@ -39,7 +39,7 @@ namespace SerialCommands {
     CmdCallback<6> cmdCallbacks;
     CmdParser cmdParser;
     CmdBuffer<256> cmdBuffer;
-	bool cmdFromRemote = false;
+    bool cmdFromRemote = false;
 
 
 	bool lengthCheck (const char* const text, unsigned int length, const char* const cmd, const char* const name)


TheDevMinerTV (approved):
Great, more network stuff for me to refactor in impl/feature-retrieval

LGTM

src/logging/Logger.h
```diff
@@ -5,6 +5,12 @@
 #include "debug.h"
 #include <Arduino.h>
 
+// Avoid circular dependancy on Network - Logger - Network
+#include <WiFiClient.h>
+namespace Network {
```
TheDevMinerTV, line 10: Actually, we have a file for this: `GlobalVars.h`

src/network/network.cpp
```diff
@@ -22,7 +22,14 @@
 */
 #include "network.h"
 
-bool lastWifiConnected = false;
+namespace Network {
+    WiFiServer rcmdServer(23);
```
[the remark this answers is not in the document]
nekomona: Like give it a config switch in `defines.h`?

src/logging/Logger.h
nekomona: Most of the modules in `GlobalVars.h` reference Logger in some way, so including `GlobalVars.h` into `Logger.h` would raise error about using Logger before defining them. Couldn't throw network into `Logger.cpp` as there's a template function using network.

src/network/network.cpp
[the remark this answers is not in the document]
TheDevMinerTV: Yeah

TheDevMinerTV:
It'd be cool if you could rebase this onto #256

nekomona:
pain^2

src/logging/Logger.h
TheDevMinerTV: I see

TheDevMinerTV (approved):
LGTM, would be nice if it were rebased against the network rewrite stuff tho

src/logging/RemoteLogHelper.h
TheDevMinerTV, line 1: please format the files using the clang-format file ^^

src/logging/RemoteLogHelper.cpp
TheDevMinerTV, line 1: please format the files using the clang-format file ^^

src/network/remotecmd.h
TheDevMinerTV, line 1: please format the files using the clang-format file ^^

src/network/remotecmd.cpp
TheDevMinerTV, line 1: please format the files using the clang-format file ^^

TheDevMinerTV (changes requested):

src/network/remotecmd.cpp
nekomona: All new files are formatted -w-/

TheDevMinerTV (approved):

nekomona:
IP check is added, so that remote command would only accept connection from the IP of connected SlimeVR Server. When the server is not connected, any IP could connect to the remote command, but will not receive any response, and only accept the `SET WIFI` command for future SoftAP-based provisioning.

ImUrX:
commit history is a little broken on this and #262 :sweat:

Eirenliel:
We squash it so it doesn't matter..

nekomona:
Rebased to main after v0.4.0 and tested on ESP32-S3, remote commands and server IP checks are looking good.

src/serial/serialcommands.cpp
```diff
@@ -147,6 +148,10 @@ namespace SerialCommands {
     }
 
     void cmdGet(CmdParser * parser) {
+#if USE_REMOTE_COMMAND && ALLOW_REMOTE_WIFI_PROV
+		if (cmdFromRemote && !networkConnection.isConnected()) return;
+#endif
```
ButterscotchV, line 153: This is used multiple times throughout this file, but it's not applied to every command? Why not just put it at the parser before the command functions are called? Also why is it on get and such? Isn't WiFi prov just for setting the WiFi?

loucass003:
As much as i would love this feature. There might be security concerns to think about. As sending commands over wifi would allow the usage of the reboot command remotely.

Rebooting will make the OTA reset and allow for flashing new firmwares without the user interaction. 

As is, this PR is **dangerous.**

If we could find a way to secure the connection when sending those commands it would be a super cool feature

loucass003 (changes requested):
Too dangerous as is

ButterscotchV:
Can we just cut out the sending commands part and focus on the receiving bit for now instead? I think it would be interesting to handle sending but there are definitely considerations to be made about security there

src/network/featureflags.h
```diff
@@ -73,8 +74,8 @@ class FirmwareFeatures {
 public:
     enum EFirmwareFeatureFlags: uint32_t {
         // EXAMPLE_FEATURE,
+		REMOTE_COMMAND = 0,
```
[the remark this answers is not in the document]
nekomona: It's just from that the first version of remote command is written before WiFi scanning

src/serial/serialcommands.cpp
nekomona: The idea here is to only allow `SET` and block other commands in the provisioning phase, before connected to a slime server. The repetition is from that the command is only known to be a `SET` after parsing, which is processed inside cmdParser.

nekomona:
> Can we just cut out the sending commands part and focus on the receiving bit for now instead? I think it would be interesting to handle sending but there are definitely considerations to be made about security there

I think we could let the command part to be an define option disabled by default, and probably reject `REBOOT` and `FACTORY RESET` from remote completely.

ImUrX:
> I think we could let the command part to be an define option disabled by default, and probably reject `REBOOT` and `FACTORY RESET` from remote completely.

sounds good

PR #256, opened 2023-06-30
Rewrite Network and UdpClient into a classes

TheDevMinerTV:
I've done the long needed UDPClient and Network namespace to class rewrite ^^

PR #262, opened 2023-07-12
Add `GET WIFISCAN` and base64 WiFi credential commands

nekomona:
This PR add `GET WIFISCAN` command to initiate a blocking scan, retrieving what hotspot could be seen by the trackers during provisioning. Could avoid connecting to a 5G hotspot, misspelling, and might help future fancy provisioning.

The result would be printed in the format below:
```
[INFO ] [SerialCommands] [WSCAN] Scanning for WiFi networks...
[INFO ] [SerialCommands] [WSCAN] Found <Count> networks:
[INFO ] [SerialCommands] [WSCAN] 0:\t<SSID0 Length>\t<SSID0>\t(<RSSI0>)\t<OPEN/PASS>
[INFO ] [SerialCommands] [WSCAN] 1:\t<SSID1 Length>\t<SSID1>\t(<RSSI1>)\t<OPEN/PASS>
```

In addition, since `CmdParser` couldn't deal with double quotes in SSID / Password, a base64 version for serial provisioning is added (`SET BWIFI <B64SSID> <B64PASS>`). Everything is the same except encoded in base64 to avoid double quote in the command.

Currently blocked by #256


ImUrX:
btw is it necessary for it to be blocking? I have no idea if you can just print it when you get the scan data

nekomona:
I've tried the non-blocking one during the first commit, but it always returns as failed. There should be somewhere buggy but can't locate them. The blocking scan will timeout server connection but reconnects once finished, and it won't break WiFi connection. Since we only need to scan during provisioning, imo the blocking one would be enough.

Eirenliel (approved):

src/serial/serialcommands.cpp
```diff
@@ -37,19 +38,37 @@ namespace SerialCommands {
...
+			} else if (parser->equalCmdParam(1, "BWIFI")) {
+				if(parser->getParamCount() < 3) {
+					logger.error("CMD SET BWIFI ERROR: Too few arguments");
+					logger.info("Syntax: SET BWIFI <B64SSID> <B64PASSWORD>");
+				} else {
+					char ssid[33];
+					char pass[65];
+					const char * b64ssid = parser->getCmdParam(2);
+					const char * b64pass = parser->getCmdParam(3);
+					base64_decode_chars(b64ssid, strlen_P(b64ssid), ssid);
```
unlogisch04, line 62: Is this conversation length save?  so ssid will not be writen longer that it is?

src/serial/serialcommands.cpp
```diff
@@ -37,19 +38,37 @@ namespace SerialCommands {
...
+					base64_decode_chars(b64pass, strlen_P(b64pass), pass);
```
unlogisch04, line 63: Same here

src/serial/serialcommands.cpp
```diff
@@ -153,6 +172,32 @@ namespace SerialCommands {
...
             }
         }
+
+        if (parser->equalCmdParam(1, "WIFISCAN")) {
+			logger.info("[WSCAN] Scanning for WiFi networks...");
+
+			// Scan would fail if connecting, stop connecting before scan
+			if (WiFi.status() != WL_CONNECTED) WiFi.disconnect();
+
+			WiFi.scanNetworks();
```
unlogisch04, line 182: I did try it on ESP8266:
It did not work for me after a short time after boot, after the line
`[INFO ] [WiFiProvisioning] SmartConfig started` it did return everytime:
`[INFO ] [SerialCommands] [WSCAN] Scan failed!`

i would also love to have this function async, because it will probably mess up some IMUs because they need to fetch data or they may timeout.

unlogisch04 (commented):
There some stuff that need to be locked at.

PR #263, opened 2023-07-23
Implement packet bundling

0forks:
Companion PR: https://github.com/SlimeVR/SlimeVR-Server/pull/794

ESP8266 does not support frame aggregation natively in its networking implementation, this PR adds it on application level to decrease airtime by 50-70%.
- New packet 22 (0x16): feature flags - resizable bitfield (uint16 length prefix + uint8[length]);
- New packet 100 (0x64): bundle - contains multiple type- (uint32) and length- (uint16) prefixed packets.

In case of single sensor, sends rotation and acceleration packets (and everything produced by `Sensor::sendData`) as one.
In case of multiple sensors, by default attempts to wait for data from all sensors with a timeout of 10ms (send at no less than 100Hz).


src/network/connection.cpp
```diff
@@ -89,31 +101,80 @@ bool Connection::endPacket() {
...
+	
+	return false;
+}
+
+bool Connection::endBundle() {
+	m_IsBundle = false;
+
+	if (!m_ServerFeatures.has(ServerFeatures::PROTOCOL_BUNDLE_SUPPORT)) {
+		return false;
+	}
```
Eirenliel, line 128:
```suggestion
	if(!m_IsBundle)
	    return;
```

Eirenliel (approved):

src/network/connection.cpp
nekomona:
```suggestion
	if (!m_IsBundle) {
		return false;
	}
	
	m_IsBundle = false;
```
Still need to clear the flag before ending that packet

nekomona (approved):
LGTM -w-/

Eirenliel (approved):

Issue #245, opened 2023-05-21
AUX BMI160 tracker doesn't wake up from rest on 0.3.3


KonradOlech:
Seems like when using 0.3.3 release aux trackers on mine devices stop reporting its position to SlimeVR server after some time of inactivity. It looks like there is some problem with waking up trackers from sleep.
Returning to 0.3.2 resolves this problem so it's not hardware issue.

I have custom board with ESP32-C3-MINI-1

PR #265, opened 2023-07-26
ESP32C3 stability improvements

l0ud:
This PR contains my recent findings when enabling ESP32C3 board. With this changes the following issues are fixed:
- BMI160 shaky/not functional/completely drifty before calibration (because of uninitialized memory)
- BMI160 being broken after some amount of time, especially on rest ( may help with https://github.com/SlimeVR/SlimeVR-Tracker-ESP/issues/245 ) when temperature calibration is enabled (default).
- (Possibly) other sensors with calibration stored in tracker being broken before performing it.
- Trackers with AUX IMU compiled but not connected making server going crazy with OK/Error status changing every second (workaround for https://github.com/espressif/arduino-esp32/issues/8454 )

### DIFF
--- a/lib/i2cscan/i2cscan.cpp
+++ b/lib/i2cscan/i2cscan.cpp
@@ -20,10 +20,10 @@ namespace I2CSCAN
 {
 
     uint8_t pickDevice(uint8_t addr1, uint8_t addr2, bool scanIfNotFound) {
-        if(I2CSCAN::isI2CExist(addr1)) {
+        if(I2CSCAN::hasDevOnBus(addr1)) {
             return addr1;
         }
-        if(I2CSCAN::isI2CExist(addr2)) {
+        if(I2CSCAN::hasDevOnBus(addr2)) {
             return addr2;
         }
         if (scanIfNotFound) {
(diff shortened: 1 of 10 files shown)


lib/i2cscan/i2cscan.cpp
```diff
@@ -111,8 +112,16 @@ namespace I2CSCAN
...
-        byte error = Wire.endTransmission();
+        byte error;
+#if ESP32C3
+        do {
+#endif
+            Wire.beginTransmission(addr);
+            error = Wire.endTransmission();
+#if ESP32C3
+        }
+        while (error == 5);
```
nekomona, line 123: Do we need a timeout here if the bus is locked up like SCL tied to ground?

nekomona (approved):
Would like to check if the timeout matters, others are LGTM

lib/i2cscan/i2cscan.cpp
l0ud: You were right, it can freeze on shorted bus. Added try counter. Tested with bmi160 and it require only one try.

nekomona (approved):
LGTM now -w-/

Vyolex (approved):

Eirenliel (approved):

PR #266, opened 2023-07-30
Fix sending ErroneousSensor if not found

0forks:
Since #249 the firmware has a regression in sensor assignment logic and sets an `ErroneousSensor` instead of leaving an `EmptySensor` as is.

If there are multiple IMUs in defines (main + aux) and some of them are not connected (aux), it sends an `ErroneousSensor`.
Server logic for UDP tracker statuses also seems weird - it briefly shows an error, then gets overwritten with OK.
image

After the change in this PR the extra tracker no longer appears.

For reference old aux logic that used to scan all addresses:
https://github.com/SlimeVR/SlimeVR-Tracker-ESP/blob/ed74944551cc7a635ce0ef2d2c66637a021a157b/src/sensors/SensorManager.cpp#L88-L97

Also logging below uses `sensorId+1` style of sensor numbering, so the new "not found" line uses the same.
https://github.com/SlimeVR/SlimeVR-Tracker-ESP/blob/e33d3e448ef74301876eaf72a9c0dcd34bb32d4b/src/sensors/SensorManager.cpp#L152-L156

### DIFF
--- a/src/sensors/SensorManager.cpp
+++ b/src/sensors/SensorManager.cpp
@@ -58,9 +58,10 @@ namespace SlimeVR
             swapI2C(sclPin, sdaPin);
 
             if (I2CSCAN::isI2CExist(address)) {
-                m_Logger.trace("IMU %d found at address 0x%02X", sensorID, address);
+                m_Logger.trace("Sensor %d found at address 0x%02X", sensorID + 1, address);
             } else {
-                sensor = new ErroneousSensor(sensorID, imuType);
+                m_Logger.debug("Sensor %d not found at address 0x%02X", sensorID + 1, address);
+                sensor = new EmptySensor(sensorID);
                 return sensor;
             }
 


Vyolex (approved):

l0ud:
This will make ErroneousSensor not being set even on valid case, like first IMU not found. I would consider first sensor as mandatory to (for example) report badly soldered BMI160 as errored.
On my main branch I have a little more extended patch you could consider: https://github.com/l0ud/SlimeVR-Tracker-ESP-BMI270/commit/5c29b04a3bd96ef4817a15d8c1e64b0f37da63c4

0forks:
People have mixed primary IMU addresses, so wouldn't that leave an extra sensor when primary is id=1 and aux isn't connected?

I think for now we should keep how it used to work before all the refactors - never send sensor state for IMUs that aren't there. It still prints a serial message, it's just a different message (and IMO more correct) - "not found" instead of "failed to initialize".

Then if we want to show a warning for a connected device with 0 trackers, it might be better done on the server/UI side and without any extra configuration.

l0ud:
> People have mixed primary IMU addresses, so wouldn't that leave an extra sensor when primary is id=1 and aux isn't connected?
I'm aware of website showing that wiring for bno, but it's not going to be supported in current state of FW anyway - interrupt assignment problem.
In other cases people are instructed to set lower address to main imu in basically all schematics (I'm aware of).

I don't agree we should keep old behavior for invalid trackers. Badly soldered/completely DOA BMIs show no sensors at all and we had many questions form users like "Tracker connected successfully but not visible, why". In current state of FW and server they show error. It's better IMO. Btw, in my proposal you can set main to non-mandatory if you really want (it's just mandatory by default).

> I think for now we should keep how it used to work before all the refactors
I think it's impossible to do cleanly with current state of firmware. At least bno dynamic interrupt assignment would be VERY messy and kinda against idea of sensor descriptors.
I'm a little biased though, I didn't like "magic" assignment in SensorManager form the very beginning.

Vyolex:
I see both points being made.
It depends if we want to trust the defines as ground truth, which we should imo since there is no other way to verify what should be there. The issue is that on the official design the extensions are "hot-swappable", so they would somehow need to be marked as optional as you suggested. In this case a faulty extension that does not show up for whatever reason will be ignored however. 

Ideally we'd have users put the exact specifications that they will use in the defines, but in case of the official slimes this won't be the case. Because of how the official slimes work I would suggest either restoring old behavior for now or keeping new behavior with extensions/aux trackers marked as optional. 

Neither of these are perfect solutions.

Eirenliel (approved):

PR #271, opened 2023-08-02
Allow more than 1 server flags packet per connection

0forks:
Allow updating server feature flags in case we want to make them toggleable in future.

### DIFF
--- a/src/network/connection.cpp
+++ b/src/network/connection.cpp
@@ -696,19 +696,18 @@ void Connection::update() {
 				break;
 			}
 
-			uint32_t flagsLength = len - 12;
-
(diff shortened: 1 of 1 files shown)


TheDevMinerTV (approved):

ImUrX:
how hard would you think is to make server have feature flags be different per tracker, my original idea was for disabling mags at runtime

Eirenliel:
> how hard would you think is to make server have feature flags be different per tracker, my original idea was for disabling mags at runtime

this is not a feature. "feature" is what server and/or tracker and/or sensor supports. what you're talking is config and should be treated as such.

i also think that this PR isn't needed too, but it's a trivial change so i'll merge it

PR #274, opened 2023-08-10
Add new packet recieve type: LED Blink Offset

nullstalgia:
Progresses (but does *not* close) [issue 301 of the SlimeVR Server](https://github.com/SlimeVR/SlimeVR-Server/issues/301).

This allows the server to send a packet to the tracker with a uint32_t value that contains the delay before it should (re)start the normal blinking loop.

Used for syncing the blinks together, or set up an order in which trackers should blink with incrementing offsets.


Eirenliel:
Is this *really* necessary...

nullstalgia:
*Absolutely necessary?* No.

But it's something that myself and others have discussed in the discord before as a cute nice-to-have, and it's not taking up massive amount of resources on either the server or tracker side. I opened the issue in November of 2022 and since no one else took up the mantle (or shut it down), myself and Uriel were discussing it after another person was asking about it. We decided to make it a reality together, since I knew the ESP side better than the Kotlin side, and visa versa.

PR #280, opened 2023-09-07
Fix MPU9250 calibration matrix multiplication

furrycoding:
Split applying calibration values into 2 loops - first the bias, then the matrix(like for BMI160)
 If done in a single loop the matrix multiplication uses uninitialized `temp[]` values, which is wrong


furrycoding:
I have realized #246 exists, which is basically the same thing
Always remember to check existing pulls first -w-

unlogisch04:
The PR #246 got merged. So feel free to test the main.

PR #281, opened 2023-09-08
BMI160 sens calibration

wigwagwent:
Instead of using the defines_sensitivity.h to set the gyro sensitivity, as part of the default calibration you spin the tracker 2 times on each axis in the yaw direction and it calculates the sensitivity for you and stores it in the calibration on the esp.


unlogisch04:
Is this pr still needed? looking at #322 ?

Eirenliel:
Superceeded by SoftFusion

PR #283, opened 2023-09-19
Fix mahony/madgwick updates

0forks:
It wasn't passing all gyro samples to fusion if acceleration update rate is lower, resulting in broken rotation, also `magExist` condition was inverted.
Only useful for testing though.


unlogisch04 (approved):
LGTM

Eirenliel (approved):

PR #288, opened 2023-09-20
Report IMU errors

l0ud:
This is changing default behavior of Sensors. If sensor failed to initialize (and has working=false) it will report error, not offline.
I added exception to Empty sensor, not to generate errors from disconnected (and not mandatory) aux imus.

I also made sensor manager set IMU_ERROR flag to statusManager. This way user will see 5 blinking leds on imu error. So far only bno sensor was setting that flag.


unlogisch04 (approved):
LGTM

Eirenliel (approved):

PR #290, opened 2023-09-22
fix Imu icm42688

sctanf:


unlogisch04:
does not build
 #include "network/network.h"

sctanf:
ok done updating everything i think; there still is an issue with calibrating maybe but i cannot test it. same with axes order

unlogisch04 (approved):
LGTM (it build)

Eirenliel (approved):

PR #291, opened 2023-09-22
MPU-6050 No-DMP mode

furrycoding:
Adds a new type of sensor that uses MPU-6050, but without the DMP(sensor fusion in software)
 * Pros:
   * No need to keep MPUs sill/flat on the ground when powering on
 * Cons:
   * Potentially higher drift
   * Manual calibration is basically required

Uses the same sensor fusion as BMI160


furrycoding:
TODO: Verify that rest detection for TCAL actually works(MPUs accel/gyro tend to be quite noisy)

I need to find clean way to modify rest detection config however

furrycoding:
Rebased the branch to 0.4.0
Doing some testing right now, but eveything seems to be working well

src/consts.h
```diff
@@ -35,6 +35,10 @@
 #define IMU_BMI160 8
 #define IMU_ICM20948 9
 #define IMU_ICM42688 10
+
+#define IMU_MPU6500_NODMP 102
+#define IMU_MPU6050_NODMP 106
```
ButterscotchV, line 40: Idk if I like these being defined this way... I think I'd prefer it just following the incrementing ID order.

README.md
```diff
@@ -18,6 +18,13 @@ The following IMUs and their corresponding `IMU` values are supported by the fir
 * MPU-6500 (IMU_MPU6500) & MPU-6050 (IMU_MPU6050)
   * Using internal DMP to fuse Gyroscope and Accelerometer. Can drift substantially.
   * NOTE: Currently the MPU will auto calibrate when powered on. You *must* place it on the ground and *DO NOT* move it until calibration is complete (for a few seconds). **LED on the ESP will blink 5 times after calibration is over.**
+* No-DMP MPU-6500 (IMU_MPU6500_NODMP) & MPU-6050 (IMU_MPU6050_NODMP)
+  * Works with the same IMU, but doesn't use the DMP. Can still drift substantially(perhaps even more, depends on calibration, needs more testing)
+  * Enable by changing IMU type(append `_NODMP` at the end)
+  * For decent results, you **have** to perform gyro/accel calibration(Ideally at the same temperature the tracker would be operating at)
+  * Change calibration config in `defines_mpu6050.h`
+  * The calibration process itself is basically identical to that of BMI160, so refer to it's section for details
+  * You can also perform temperature calibration for better results in variable temperature(slowly heat up the tracker from below 10 degrees C to above 45, keeping it very still)
```
ButterscotchV, line 27:
```suggestion
* No-DMP MPU-6500 (IMU_MPU6500_NODMP) & MPU-6050 (IMU_MPU6050_NODMP)
  * Works with the same IMU, but doesn't use the DMP. Can still drift substantially (perhaps even more, depends on calibration, needs more testing).
  * Enable by changing IMU type (append `_NODMP` at the end)
  * For decent results, you **must** perform gyro/accel calibration (ideally at the same temperature the tracker would be operating at)
  * Modify calibration config in `defines_mpu6050.h`
  * The calibration process itself is basically identical to that of BMI160, so refer to its section for details
  * You can also perform temperature calibration for better results in variable temperature (slowly heat up the tracker from below 10 degrees C to above 45, keeping it very still)
```

README.md
```diff
@@ -51,7 +58,7 @@ Firmware can work with both ESP8266 and ESP32. Please edit `defines.h` and set y
   * You don't have to calibrate next time you power it on, calibration values will be saved for the next use
 
 ### BMI160
-
+#### (or MPU6050 in no-DMP mode)
```
ButterscotchV, line 61: This should probably just be one line
```markdown
### BMI160 or MPU6050 in no-DMP mode
```

wb1016:
it is working on ESP8266 and ESP32C3. one thing that i would change is display name of IMU.
asdfasdfsfadsdafasdcasdf

wb1016:
works on ESP8266 and ESP32C3. one thing i would suggest to change is display name of IMU.
asdfasdfsfadsdafasdcasdf

furrycoding:
> one thing i would suggest to change is display name of IMU. asdfasdfsfadsdafasdcasdf

Would need to add a new IMU type in the server, SolarXR, and finally in the UI. I do agree that it'd be nice to know what mode the MPU is in, but using the IMU type for this would require too many changes

furrycoding:
Rebased onto latest main

src/consts.h
furrycoding: Using IDs 13 & 14 now(skipping ahead a bit to account for #296 and #308)

unlogisch04:
I'm not sure if there is still need for this PR, as the softfusion branch has a MPU6050 with softfusion available.

furrycoding:
This branch still has temperature calibration, which sfusion currently lacks. However, that'll soon likely change. So there's no point in this PR anymore. (However the branch itself has some use)

PR #293, opened 2023-09-25
Fix WiFiscan not working when not connected

unlogisch04:
also changed tab to 4 space, and a 1 line if to a 3 line if


ImUrX:
you should have made the tab change in a different commit, diff is not readable qwq
And the [editorconfig](https://github.com/SlimeVR/SlimeVR-Tracker-ESP/blob/main/.editorconfig) specifies tab instead of space

unlogisch04:
Ok i had to search a littlebit what should be.
Seems like to be tab the correct one for future:
https://discord.com/channels/817184208525983775/959133879682342962/1089772491486011465
https://github.com/SlimeVR/SlimeVR-Tracker-ESP/pull/235

unlogisch04:
Ok, i did not have "EditorConfig for VS Code" installed so the rules where not applied on my code

src/serial/serialcommands.cpp
```diff
@@ -196,7 +201,10 @@ namespace SerialCommands {
 			}
 
 			// Restore conencting state
-			if (WiFi.status() != WL_CONNECTED) WiFi.begin();
+			if (WiFi.status() != WL_CONNECTED) {
+				WiFi.begin();
+			}
+
```
TheDevMinerTV, line 207:
```suggestion
			}
```

src/serial/serialcommands.cpp
```diff
@@ -196,7 +201,10 @@ namespace SerialCommands {
...
         }
```
TheDevMinerTV, line 208:
```suggestion
			if (WiFi.status() != WL_CONNECTED) {
				WiFi.begin();
			}
        }
```

TheDevMinerTV (approved):
LGTM

TheDevMinerTV (approved):
LGTM

TheDevMinerTV:
ignore that I reviewed twice 😱

PR #294, opened 2023-10-01
Clean up packet bundles and feature flags

TheDevMinerTV:


src/network/connection.h
```diff
@@ -107,28 +107,25 @@ class Connection {
...
+	size_t write(const uint8_t* buffer, size_t size);
 
 	bool sendPacketType(uint8_t type);
 	bool sendPacketNumber();
 	bool sendFloat(float f);
 	bool sendByte(uint8_t c);
-	bool sendShort(int16_t i);
+	bool sendU16(uint16_t c);
 	bool sendInt(int32_t i);
 	bool sendLong(int64_t l);
```
unlogisch04, line 133: i think this should be uint64_t as the m_PacketNumber is the only thing that is send with that. see above

src/network/connection.cpp
```diff
@@ -118,7 +118,8 @@ bool Connection::beginBundle() {
 	memset(m_BundledPacket, 0, sizeof(m_BundledPacket));
 	m_BundlePacketInnerCount = 0;
 
-	MUST_TRANSFER_BOOL(m_ServerFeatures.has(ServerFeatures::PROTOCOL_BUNDLE_SUPPORT));
+	MUST_TRANSFER_BOOL(m_ServerFeatures.has(EServerFeatureFlags::PROTOCOL_BUNDLE_SUPPORT
+	));
```
unlogisch04, line 122: Why?

src/network/connection.cpp
```diff
@@ -197,13 +207,7 @@ bool Connection::sendShortString(const char* str) {
...
 }
 
-bool Connection::sendPacketType(uint8_t type) {
-	MUST_TRANSFER_BOOL(sendByte(0));
-	MUST_TRANSFER_BOOL(sendByte(0));
-	MUST_TRANSFER_BOOL(sendByte(0));
-
-	return sendByte(type);
-}
+bool Connection::sendPacketType(uint8_t type) { return sendInt(type); }
```
unlogisch04, line 211: Not sure if that is what is intended. is PacketType 3Byte Reserve and then 1Byte of Data, or is it 4Byte 1 Number?

And i think no 1 line functions. We prefer the 3 line if?

src/network/featureflags.h
```diff
@@ -1,103 +1,99 @@
...
-
-private:
-    bool m_Available = false;
-
-    uint8_t m_Flags[static_cast<uint32_t>(EServerFeatureFlags::BITS_TOTAL) / 8 + 1];
+#include <cinttypes>
+#include <cstring>
+#include <unordered_map>
+
+// I hate C++11 - they fixed this in C++14, but our compilers are old as the iceage
```
unlogisch04, line 33: hm look platform.ini is there not gcc17 active? `-std=gnu++17`

unlogisch04 (commented):
I'm not sure about the changes in featureflags. I can't review that. That template/class is something that i understand too less.

The following should be reviewed, i could not read the code from the other side so i dont know wht it expects:
sendPacketType
sendLong (uint64_t or int64_t)

0forks:
The new packet bundling logic is not working as intended, sends empty bundles flooding the network when there's no data written between `beginBundle()` and `endBundle()`. Idle PPS is expected be around 5-10, I'm observing close to 1000.

image

I mentioned previously that we can't write the bundle header to the UDP library buffer from `beginBundle()` because an API to reset it doesn't exist. I suggest reverting to the original logic or making a dynamic buffer (or just large enough: `64 * MAX_IMU_COUNT`) to hold the whole bundle, which is slightly less efficient. Either way it'll look messy I think.

https://github.com/SlimeVR/SlimeVR-Tracker-ESP/blob/dd2156669d48e38720729d146e1021da9e1a94b3/src/network/connection.cpp#L117-L128

We could alternatively fix it by changing checks in the `Sensor`/`SensorManager`, but that would be less flexible, for example, if we want to wrap more than just `sensor->sendData()`.

Maybe also worth moving the list of default firmware flags back into `featureflags.h` so it's more obvious that they can be marked as enabled when adding new features.

https://github.com/SlimeVR/SlimeVR-Tracker-ESP/blob/dd2156669d48e38720729d146e1021da9e1a94b3/src/network/connection.h#L173-L175

src/network/featureflags.h
TheDevMinerTV: not all boards have c++ 17 support yet

src/network/connection.cpp
TheDevMinerTV: This is not an if, also, this is just what the formatter does.

src/network/connection.cpp
```diff
@@ -146,9 +146,7 @@ size_t Connection::write(const uint8_t *buffer, size_t size) {
 	return m_UDP.write(buffer, size);
 }
 
-size_t Connection::write(uint8_t byte) {
-	return write(&byte, 1);
-}
+size_t Connection::write(uint8_t byte) { return write(&byte, 1); }
```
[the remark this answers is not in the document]
TheDevMinerTV: just what the formatter did

src/network/connection.h
TheDevMinerTV: I think you're looking at the wrong thing? I only changed `sendShort()`, not `sendLong()`.

src/network/connection.cpp
TheDevMinerTV: wdym "Why?"?

src/network/connection.cpp
```diff
@@ -158,19 +156,19 @@ bool Connection::sendFloat(float f) {
 
 bool Connection::sendByte(uint8_t c) { return write(&c, 1) != 0; }
 
-bool Connection::sendShort(uint16_t i) {
+bool Connection::sendShort(int16_t i) {
 	convert_to_chars(i, m_Buf);
```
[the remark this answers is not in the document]
TheDevMinerTV: > ::sendShort is used in endPacket for innerPacketSize that is a uint32_t but it only should send 2bytes?

`innerPacketSize` is a `uint16_t`

> ::sendInt is used to send different board infos

that's how the protocol works

> ::sendLong is used to send a uint64_t for the packetnummer.

You're right, I've renamed the function to what it actually does

TheDevMinerTV:
I've moved the list of active features next to where they are declared.

Next time please use review comments ^^

TheDevMinerTV:
> The new packet bundling logic is not working as intended, sends empty bundles flooding the network when there's no data written between beginBundle() and endBundle().

I fixed this now, it how has a buffer of the size of `MAX_IMU_COUNT * 53` (53 being the size of the inspection packet).

TheDevMinerTV:
@0forks @unlogisch04 @Eirenliel please review

src/network/connection.cpp
unlogisch04: to turncat the "));" i know autoformater.

src/network/featureflags.h
```diff
@@ -1,103 +1,102 @@
...
+struct EnumClassHash {
+	template <typename T>
```
unlogisch04, line 35: I do not like the template here. As we do not save any programming bytes with it. Also the code understanding is more difficult in my meaning. I think that's a possible improvement for the future.

unlogisch04 (approved):
For me it seems to work. But i did not test it with trackers.
For ICM20948 there was a fix needed as they had a internal delay to not send too much updates.
For me the featureflags is something that could be improved more in the meaning more simple as 1 class / function that is seeded with different enums.

unlogisch04:
I did check the performance on that PR. It seems todo the relevant function (has) is about 8 times slower: It only gets called currently at each packet once. so about 100-120times a sec.

FeatureFlag1 = the new function
FeatureFlag2 = existing function
FeatureFlag3 = Raw && 2 uint8_t > 0
```
FeatureFlag1 Setuptime: 9 us
FeatureFlag1 100000     reads: 81500 us
FeatureFlag1 1      avg reads: 0.815 us
FeatureFlag2 Setuptime: 1 us
FeatureFlag2 100000     reads: 11258 us
FeatureFlag2 1      avg reads: 0.113 us
FeatureFlag3 Setuptime: 4 us
FeatureFlag3 100000     reads: 10049 us
FeatureFlag3 1      avg reads: 0.100 us
```

PR #301, opened 2023-10-31
feat: allow connecting to one specific server IP without broadcasting

TheDevMinerTV:
as @ButterscotchV requested, two flags to implicitly disable broadcast spamming the network. :)

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -35,6 +35,9 @@ build_flags =
 ;  -DWIFI_STATIC_GATEWAY=192,168,XXX,XXX
 ;  -DWIFI_STATIC_SUBNET=255,255,255,0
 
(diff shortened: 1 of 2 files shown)


unlogisch04:
I think for a test it is good enough. 
But not for final. Because the server ip-address could be different each time. The main question for that feature is how the tracker gets used. So i would see that more in a config file and a serial config option.

unlogisch04 (approved):
Tested on ESP8266 and ESP32C3.
Could connect over a router from Tracker IP 192.168.2.200 to the server 192.168.1.100

Even when i think it would be better to have a serial config for this.
But to have something is better to have nothing.

ButterscotchV (approved):
Not tested, but looks fine (I don't have a very good way of testing).

PR #305, opened 2023-12-01
Fix rest detection timescale

wigwagwent:
Sensor fusion expects the time to be in seconds while rest detection expects it to be in microseconds. This makes is so when the update sensor fusion is called the rest detection now gets the time in micros. Also linear acceleration was missing parentheses giving wrong results (only lsm6dsv uses this to my knowledge)

### DIFF
--- a/src/motionprocessing/RestDetection.h
+++ b/src/motionprocessing/RestDetection.h
@@ -20,14 +20,14 @@
 struct RestDetectionParams {
     sensor_real_t biasClip;
     sensor_real_t biasSigmaRest;
-    uint32_t restMinTimeMicros;
+    uint32_t restMinTime;
     sensor_real_t restFilterTau;
     sensor_real_t restThGyr;
     sensor_real_t restThAcc;
     RestDetectionParams():
         biasClip(2.0f),
         biasSigmaRest(0.03f),
-        restMinTimeMicros(1.5 * 1e6),
+        restMinTime(1.5),
         restFilterTau(0.5f),
         restThGyr(2.0f),
         restThAcc(0.5f)
@@ -103,7 +103,7 @@ public:
     }
 #endif
 
-    void updateGyr(uint32_t dtMicros, sensor_real_t gyr[3]) {
+    void updateGyr(sensor_real_t gyr[3]) {
 #ifdef REST_DETECTION_DISABLE_LPF
         gyrLastSquaredDeviation =
             square(gyr[0] - lastSample.gyr[0]) +
@@ -114,7 +114,7 @@ public:
         if (gyrLastSquaredDeviation >= square(params.restThGyr*sensor_real_t(M_PI/180.0))
                 || fabs(lastSample.gyr[0]) > biasClip || fabs(lastSample.gyr[1]) > biasClip
                 || fabs(lastSample.gyr[2]) > biasClip) {
-            restTimeMicros = 0;
+            restTime = 0;
             restDetected = false;
         }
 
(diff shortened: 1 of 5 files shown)


ButterscotchV:
The time scale can probably be moved to `RestDetection.h` can't it? I think it's odd to have it different, otherwise this does look good I think?

src/sensors/SensorFusion.cpp
```diff
@@ -173,9 +173,9 @@ namespace SlimeVR
         
         void SensorFusion::calcLinearAcc(const sensor_real_t accin[3], const sensor_real_t gravVec[3], sensor_real_t accout[3])
         {
-            accout[0] = accin[0] - gravVec[0] * CONST_EARTH_GRAVITY;
-            accout[1] = accin[1] - gravVec[1] * CONST_EARTH_GRAVITY;
-            accout[2] = accin[2] - gravVec[2] * CONST_EARTH_GRAVITY;
+            accout[0] = (accin[0] - gravVec[0]) * CONST_EARTH_GRAVITY;
+            accout[1] = (accin[1] - gravVec[1]) * CONST_EARTH_GRAVITY;
+            accout[2] = (accin[2] - gravVec[2]) * CONST_EARTH_GRAVITY;
```
0forks, line 178: Breaking change for every IMU other than BNO.

Vector `accin` is in m/s^2 and `gravVec` is a unit vector, if you have issues with linear acceleration in another IMU implementation, perhaps you're passing acceleration into fusion in G?

0forks (changes requested):

src/sensors/SensorFusion.cpp
wigwagwent: I did not realize I was using this wrong, I have reverted my changes and will go about this another way in the IMU specific code. I will probably change the units the IMU uses.

wigwagwent:
I changed the rest detection to be in seconds instead of in microseconds like it was before. **I am not able to test this.** The best way to test would be see if you can still calibrate a bmi160.

src/motionprocessing/RestDetection.h
```diff
@@ -20,14 +20,14 @@
 struct RestDetectionParams {
     sensor_real_t biasClip;
     sensor_real_t biasSigmaRest;
-    uint32_t restMinTimeMicros;
+    uint32_t restMinTime;
```
0forks, line 23: `uint32_t` -> `sensor_real_t`

src/motionprocessing/RestDetection.h
```diff
@@ -235,7 +235,7 @@ class RestDetection {
 private:
     RestDetectionParams params;
     bool restDetected;
-    uint32_t restTimeMicros;
+    uint32_t restTime;
```
0forks, line 238: `uint32_t` -> `sensor_real_t`

src/motionprocessing/RestDetection.h
```diff
@@ -134,13 +134,13 @@ class RestDetection {
...
                 || fabs(restLastGyrLp[2]) > biasClip) {
-            restTimeMicros = 0;
+            restTime = 0;
             restDetected = false;
         }
 #endif
     }
 
-    void updateAcc(uint32_t dtMicros, sensor_real_t acc[3]) {
+    void updateAcc(uint32_t dt, sensor_real_t acc[3]) {
```
0forks, line 143: `uint32_t` -> `sensor_real_t`

0forks (changes requested):
Tested, need to change the types to `sensor_real_t`, otherwise it all gets rounded and deltas become 0.

Also minor inconsistency with tabs/spaces, but we need to fix that across the whole codebase honestly.

Eirenliel (commented):
Update from comments

Eirenliel (approved):

PR #306, opened 2023-12-01
Set C3 default clock speed to 80MHz

wigwagwent:
This reduces the esp32c3 power consumption by around 15 ma. The platformIO version was also updated to the latest for the 32 series


unlogisch04 (commented):
There are reports that with the lower clock speed it does not work. So I'm not sure if it would be better to mark it as experimental.

Eirenliel:
This would require community testing and maybe some way for people to chose "low power mode" when building FW, because this might not work anymore with newer fusions.

PR #308, opened 2023-12-03
Support LSM6DSV IMU

wigwagwent:
This adds support for the lsm6dsv family of IMUs. They are better than bmi160 and seems comparable to the bno but need more users testing them.

This merge should probably be squashed into 1 commit.


src/consts.h
```diff
@@ -35,6 +35,7 @@
 #define IMU_BMI160 8
 #define IMU_ICM20948 9
 #define IMU_ICM42688 10
+#define IMU_LSM6DSV 12
```
ButterscotchV, line 38: This should be noted when merging, depending on the order of merging PRs this might be awkward to skip a number.

ButterscotchV (approved):

README.md
```diff
@@ -34,8 +34,15 @@ The following IMUs and their corresponding `IMU` values are supported by the fir
 * ICM-20948 (IMU_ICM20948)
   * Using fusion in internal DMP for 6Dof or 9DoF, 9DoF mode requires good magnetic environment.
   * Comment out `USE_6DOF` in `debug.h` for 9DoF mode.
+  * Experimental support!\
```
TheDevMinerTV, line 37: Something broke here, what's that `\`?

lib/LSM6DSV/LSM6DSV.cpp
```diff
@@ -0,0 +1,4080 @@
+/**
```
TheDevMinerTV, line 1: Please include a link to the original repository.

src/sensors/SensorManager.cpp
```diff
@@ -108,6 +109,12 @@ namespace SlimeVR
             case IMU_ICM42688:
                 sensor = new ICM42688Sensor(sensorID, address, rotation, sclPin, sdaPin);
                 break;
+            case IMU_LSM6DSV:
+                {
+                uint8_t intPin = extraParam;
+                sensor = new LSM6DSVSensor(sensorID, imuType, address, rotation, sclPin, sdaPin, intPin);
+                }
+                break;
```
TheDevMinerTV, line 117:
```suggestion
                sensor = new LSM6DSVSensor(sensorID, imuType, address, rotation, sclPin, sdaPin, extraParam);
                break;
```

lib/math/quat.h
```diff
@@ -72,7 +72,7 @@ class Quat {
 	Vector3 get_euler_yxz() const;
 
 	void set_euler(const Vector3& p_euler) { set_euler_yxz(p_euler); };
-	Vector3 get_euler() const { return get_euler_yxz(); };
+	Vector3 get_euler() const { return get_euler_xyz(); };
```
TheDevMinerTV, line 75: Instead of modifying `get_euler()` you should rather use the `get_euler_xyz()` in your code.

TheDevMinerTV (approved):

README.md
```diff
@@ -34,8 +34,15 @@ The following IMUs and their corresponding `IMU` values are supported by the fir
 * ICM-20948 (IMU_ICM20948)
```
TheDevMinerTV, line 34: GitHub doesn't allow me to request changes without a comment...

TheDevMinerTV (changes requested):

Eirenliel (changes requested):
Please address @TheDevMinerTV's comments

src/sensors/SensorManager.cpp
Pespiri: I would suggest against this change, as the original change follow the pattern of how other cases in this switch statement has done this

image

unlogisch04:
Is this still needed if we go for #322 ?

gorbit99:
> Is this still needed if we go for #322 ?

This probably shouldn't be preferred over sfusion, at least not without a cleanup.

unlogisch04:
I convert that to draft, as sfusion seems to be the better solution.
@wigwagwent feel free to close it when you see it the same way.

PR #310, opened 2023-12-23
ESP32: Fix ADC/battery sense logic

kounocom:
At least on ESP32-C3, default logic returns completely wrong values for voltage. This fixes it.

### DIFF
--- a/src/batterymonitor.cpp
+++ b/src/batterymonitor.cpp
@@ -79,9 +79,9 @@ void BatteryMonitor::Loop()
             #endif
             #if ESP8266 && BATTERY_MONITOR == BAT_EXTERNAL
                 voltage = ((float)analogRead(PIN_BATTERY_LEVEL)) * batteryADCMultiplier;
+            #endif
             #if ESP32 && BATTERY_MONITOR == BAT_EXTERNAL
                 voltage = ((float)analogReadMilliVolts(PIN_BATTERY_LEVEL)) / 1000 * (BATTERY_SHIELD_RESISTANCE + BATTERY_SHIELD_R1 + BATTERY_SHIELD_R2) / BATTERY_SHIELD_R1;
-            #endif
             #endif
             #if BATTERY_MONITOR == BAT_MCP3021 || BATTERY_MONITOR == BAT_INTERNAL_MCP3021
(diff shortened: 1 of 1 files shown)


gorbit99:
I could also reproduce this issue and this solved it for me perfectly

TheDevMinerTV:
This could break the normal ESP32. Please double check that the voltages are still correct on it.

kounocom:
> This could break the normal ESP32. Please double check that the voltages are still correct on it.

Confirmed working on Wroom 32 board

[http://cdn.kouno.xyz/2H6KA4So.png](url)
[http://cdn.kouno.xyz/ve4F99oE.JPEG](url)

src/batterymonitor.cpp
```diff
@@ -77,9 +77,12 @@ void BatteryMonitor::Loop()
                     }
                 }
             #endif
-            #if BATTERY_MONITOR == BAT_EXTERNAL
+            #if ESP8266 && BATTERY_MONITOR == BAT_EXTERNAL
                 voltage = ((float)analogRead(PIN_BATTERY_LEVEL)) * batteryADCMultiplier;
             #endif
+            #if ESP32 && BATTERY_MONITOR == BAT_EXTERNAL
+                voltage = ((float)analogReadMilliVolts(PIN_BATTERY_LEVEL)) / 1000 * (BATTERY_SHIELD_RESISTANCE + BATTERY_SHIELD_R1 + BATTERY_SHIELD_R2) / BATTERY_SHIELD_R1;
```
ButterscotchV, line 84: Why not just modify `batteryADCMultiplier`?

kounocom: From what I understand, the multiplier is correct but ESP32 ADC pins aren't linear and `analogRead` doesn't use the calibration that every ESP32 ships with, while `analogReadMillivolts` does

ButterscotchV: Mmm I guess, I just mean basically copying the code for the multiplier seems a little silly

kounocom: I didn't see a better way to do it without also refactoring ADC behavior for ESP8266

ButterscotchV (approved):

PR #318, opened 2024-03-23
Added ON-OFF button support

gorbit99:
This PR adds in support for momentary push button based ON-OFF functionality using the ESP's deep sleep feature. 

The code was tested on an ESP8266 (specifically a D1 mini) and an ESP32-C3 (supermini esp32c3). For the latter, attaching a push button between GND and any of the RTC GPIOs (GPIO0-5), the code should work, but for the ESP8266, the following hardware needs to be added:

kép

GPIO13 can be set to any of the safe pins (4, 5, 12, 13, 14), however the other ones are already used by the default SlimeVR setup. I have tried to make the setup work with a MOSFET instead of a transistor, but didn't manage to yet.

The on-off behaviour can be activated by defining ON_OFF_BUTTON inside defines.h. For example:

```cpp
#define ON_OFF_BUTTON D7
```

for a D1 mini.

When the tracker is on, holding down the button for 1 second puts it to deep sleep (this is debounced). When it's in deep sleep, pressing the button once will wake it up, resetting it. The hold down time can be changed by defining ON_OFF_BUTTON_HOLD_TIME_MS in defines.h.

The tracker attempts to deinitialize the imus attached to it.

While it varies, this setup should drop the power consumption to somewhere around 100-200uA.

There are some quirks with the ESP8266 implementation:
- Seemingly the tracker can't be flashed while it's off. 
- After flashing, the tracker will start in deep sleep mode

Things that need checking:
- Is it possible to make the necessary circuitry for the ESP8266 out of a MOSFET
- Are the 8266 quirks solvable
- Are the deinitialization routines for the IMUs correct

It's also theoretically possible to add in a hold time for turning on the tracker for the ESP32 variant, but since this isn't at all possible for the ESP8266, I didn't implement it.

### DIFF
--- a/lib/bmi160/BMI160.cpp
+++ b/lib/bmi160/BMI160.cpp
@@ -109,6 +109,12 @@ void BMI160::initialize(uint8_t addr,
     // I2CdevMod::writeByte(devAddr, BMI160_RA_INT_MAP_2, 0x00);
 }
 
+void BMI160::deinitialize() {
+    /* Issue a soft-reset to bring the device into a clean state */
+    setRegister(BMI160_RA_CMD, BMI160_CMD_SOFT_RESET);
+    delay(12);
+}
+
 bool BMI160::getErrReg(uint8_t* out) {
     bool ok = I2CdevMod::readByte(devAddr, BMI160_RA_ERR, buffer) >= 0;
     if (!ok) return false;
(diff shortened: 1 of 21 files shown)


kounocom:
Tested on custom esp32-c3 board, works as expected

gorbit99:
Can I get some eyes on this?

gorbit99:
Brought this code up to date with the sfusion merging, now theoretically all IMUs supported by sfusion should also be supported by this PR.

gorbit99:
@Vyolex Currently there is a weird drifting issue that happens with this branch, and I want to rewrite some of the logic a bit, to make it non-blocking, so I wouldn't bother yet.

kounocom (approved):

PR #320, opened 2024-03-28
Add threading support

nekomona:
This patch adds threading for ESP32 variants, allowing more sensors on a single ESP chip.
At current stage, threading mainly separates network packet building and networking stack from sensor operations.

Depending on #314

### DIFF
--- a/src/sensors/sensor.cpp
+++ b/src/sensors/sensor.cpp
@@ -30,33 +30,75 @@ SensorStatus Sensor::getSensorState() {
 }
 
 void Sensor::setAcceleration(Vector3 a) {
+#if ESP32
+    xSemaphoreTake(updateMutex, portMAX_DELAY);
+#endif
     acceleration = a;
     newAcceleration = true;
+#if ESP32
+    xSemaphoreGive(updateMutex);
+#endif
 }
 
 void Sensor::setFusedRotation(Quat r) {
+#if ESP32
+    xSemaphoreTake(updateMutex, portMAX_DELAY);
+#endif
     fusedRotation = r * sensorOffset;
     bool changed = OPTIMIZE_UPDATES ? !lastFusedRotationSent.equalsWithEpsilon(fusedRotation) : true;
     if (ENABLE_INSPECTION || changed) {
         newFusedRotation = true;
         lastFusedRotationSent = fusedRotation;
     }
+#if ESP32
+    xSemaphoreGive(updateMutex);
+#endif
 }
 
 void Sensor::sendData() {
+#if ESP32
+    Quat lquat;
+    xSemaphoreTake(updateMutex, portMAX_DELAY);
     if (newFusedRotation) {
         newFusedRotation = false;
+        lquat = fusedRotation;
+        xSemaphoreGive(updateMutex);
+    } else {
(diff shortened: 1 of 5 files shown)


src/sensors/sensor.h
```diff
@@ -58,8 +58,8 @@ class Sensor
     virtual void postSetup(){};
     virtual void motionLoop(){};
     virtual void sendData();
-    virtual void setAccelerationReady();
-    virtual void setFusedRotationReady();
+    virtual void setAcceleration(Vector3 a);
+    virtual void setFusedRotation(Quat r);
```
TheDevMinerTV, line 62: seems like you changed more than just adding threading, was this intended?

src/sensors/mpu6050sensor.cpp
```diff
@@ -153,8 +151,7 @@ void MPU6050Sensor::motionLoop()
 
             sfusion.updateAcc(Axyz);
 
-            acceleration = sfusion.getLinearAccVec();
-			setAccelerationReady();
+			setAcceleration(sfusion.getLinearAccVec());
```
TheDevMinerTV, line 154: same here

src/sensors/icm42688sensor.cpp
```diff
@@ -172,11 +172,8 @@ void ICM42688Sensor::motionLoop() {
     if (magExists)
         sfusion.updateMag(Mxyz);
     
-    fusedRotation = sfusion.getQuaternionQuat();
-    fusedRotation *= sensorOffset;
-    acceleration = sfusion.getLinearAccVec();
-    setFusedRotationReady();
-    setAccelerationReady();
+    setFusedRotation(sfusion.getQuaternionQuat());
+    setAcceleration(sfusion.getLinearAccVec());
```
TheDevMinerTV, line 176: same here

TheDevMinerTV (changes requested):
Please double check what changes are actually needed.

nekomona:
Rebased after #314 merging

TheDevMinerTV:
Maybe we should put this behind a `#define`?

nekomona:
> Maybe we should put this behind a `#define`?

`#define` switches are up -w-

TheDevMinerTV (approved):

src/sensors/sensor.cpp
```diff
@@ -30,33 +30,75 @@ SensorStatus Sensor::getSensorState() {
...
+        newFusedRotation = false;
+        lquat = fusedRotation;
+        xSemaphoreGive(updateMutex);
+    } else {
+        xSemaphoreGive(updateMutex);
+        return;
+    }
+
+    {
+        networkConnection.sendRotationData(sensorId, &fusedRotation, DATA_TYPE_NORMAL, calibrationAccuracy);
```
Eirenliel, line 72:
```suggestion
        networkConnection.sendRotationData(sensorId, &lquat, DATA_TYPE_NORMAL, calibrationAccuracy);
```

Shouldn't it be like that?

Also can be combined my taking lquat out of the define and having the same sendRotationData be used.

Eirenliel (changes requested):

PR #321, opened 2024-03-28
fshelper: fixed ESP8266 regression caused by abstracting FS access

l0ud:
This is hotfixing ESP8266 calibration storage which seems to be broken since #319 

isDirectory method on Dir object doesn't tell about the Dir object itself, but about the entry pointed by internal pointer (which you can move using next() to iterate). As next() is never called, isDirectory will always return false (in fact, isFile too).
Considering heavy simplifications in Esp8266 filesystem regarding directories and the fact that previous ESP8266 implementation didn't try to do any file/directory fixing, I made it ESP32 only.

### DIFF
--- a/src/FSHelper.cpp
+++ b/src/FSHelper.cpp
@@ -41,11 +41,6 @@ bool ensureDirectory(const char* directory) {
 	auto dir = LittleFS.open(directory);
 	auto isDirectory = dir.isDirectory();
 	dir.close();
-#else
-	auto dir = LittleFS.openDir(directory);
-	auto isDirectory = dir.isDirectory();
-#endif
-
 	if (!isDirectory) {
 		if (!LittleFS.remove(directory)) {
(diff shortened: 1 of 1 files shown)


TheDevMinerTV:
Thank you for catching this bug. I think that just leaving this check out from the ESP8266 code is the wrong move, as I wanted to implement that specific check for it as well. We should probably fix it "the right way".

TheDevMinerTV (approved):
LGTM

TheDevMinerTV (changes requested):
As I said in the comment: ^

Eirenliel (approved):

unlogisch04:
As this one seems to be stuck, i did open one with a fix that i tested on esp8266: https://github.com/SlimeVR/SlimeVR-Tracker-ESP/pull/328

l0ud:
Superseded by: https://github.com/SlimeVR/SlimeVR-Tracker-ESP/pull/328

PR #323, opened 2024-04-08
Add support for BQ27441 battery monitor

zakrent:
Adds support for [BQ27441](https://www.ti.com/product/BQ27441-G1) battery gauge IC.
This IC is used for example in this sparkfun dev board: https://www.sparkfun.com/products/13777 and also can be used on custom PCBs.
It uses Sparkfun library for this IC: https://github.com/sparkfun/SparkFun_BQ27441_Arduino_Library, it is MIT license.
I modified this library slightly to disable `Wire.begin()` call in it.

### DIFF
--- a/src/defines.h
+++ b/src/defines.h
@@ -52,11 +52,24 @@ IMU_DESC_ENTRY(IMU_BMP160, PRIMARY_IMU_ADDRESS_ONE, IMU_ROTATION, PIN_IMU_SCL, P
         IMU_DESC_ENTRY(SECOND_IMU, SECONDARY_IMU_ADDRESS_TWO, SECOND_IMU_ROTATION, PIN_IMU_SCL, PIN_IMU_SDA, SECONDARY_IMU_OPTIONAL, PIN_IMU_INT_2)
 #endif
 
+// Battery configuration
 // Battery monitoring options (comment to disable):
 //   BAT_EXTERNAL for ADC pin,
 //   BAT_INTERNAL for internal - can detect only low battery,
 //   BAT_MCP3021 for external ADC connected over I2C
+//   BAT_BQ27441 for external BQ27441 battery gauge IC
 #define BATTERY_MONITOR BAT_EXTERNAL
+#define batterySampleRate 10000
+#define BATTERY_LOW_VOLTAGE_DEEP_SLEEP false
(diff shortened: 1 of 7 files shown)


zakrent:
Sorry, it exploded after I rebased it but now should be fine.

src/batterymonitor.h
```diff
@@ -78,6 +78,8 @@ class BatteryMonitor {
 	float getVoltage() const { return voltage; }
 	float getLevel() const { return level; }
 
+	String getExtendedInfo();
```
gorbit99, line 81: Use std::string instead

src/serial/serialcommands.cpp
```diff
@@ -203,9 +203,10 @@ void printState() {
 		}
 	}
 	logger.info(
-		"Battery voltage: %.3f, level: %.1f%%",
+		"Battery voltage: %.3f, level: %.1f%%, %s",
```
gorbit99, line 206: this adds an unnecessary leading "," when anything other than the BQ chip is in use

src/defines.h
```diff
@@ -59,9 +59,14 @@
 //   BAT_EXTERNAL for ADC pin,
 //   BAT_INTERNAL for internal - can detect only low battery,
 //   BAT_MCP3021 for external ADC connected over I2C
+//   BAT_BQ27441 for external BQ27441 battery gauge IC
 #define BATTERY_MONITOR BAT_EXTERNAL
 #endif
 
+#if BATTERY_MONITOR == BAT_BQ27441 && I2C_SPEED > 100000
+#error BQ27441 library supports only 100 kHz I2C speed
+#endif
```
gorbit99, line 68: Not sure how well this requirement would work, the sensor reading speed is already somewhat the bottleneck with 400kHz

src/serial/serialcommands.cpp
zakrent: Fixed

src/defines.h
zakrent: You're right, for daily use I use trackers iwth single IMU but if I tried it with 2nd one it got laggy.
I improved the sparkfun lib to support i2c >100 kHz.

gorbit99 (approved):

PR #325, opened 2024-04-16
Fix enabling motion bias estimation

kounocom:
`motionBiasEstEnabled` should be `true` if `VQF_NO_MOTION_BIAS_ESTIMATION` isn't defined


unlogisch04:
I generally agree. 
So that change will enable that feature for all. So is that a improvement?
@nekomona whats your meaning, can we just add it?

kounocom:
To be clear this doesn't enable motion bias estimation - the user still needs to undefine `VQF_NO_MOTION_BIAS_ESTIMATION`, this only fixes enabling it

unlogisch04 (approved):
Looks good. No impact on not modified version.

PR #327, opened 2024-05-01
Designate all initializer clauses to fix errors with newer compilers

Spacefish:
Building SlimeVR for a ESP32-C6 (with better WiFi) i get some compiler errors like:
`src/defines_sensitivity.h:53:46: error: either all initializer clauses should be designated or none of them should be`

This MR fixes this, by just adding a designation to all initializer values.

### DIFF
--- a/src/defines_sensitivity.h
+++ b/src/defines_sensitivity.h
@@ -50,11 +50,11 @@
 struct SensitivityOffsetXYZ { const char* mac; unsigned char sensorId; double spins; double x; double y; double z; };
 const SensitivityOffsetXYZ sensitivityOffsets[] = {
     // example values
-    { "A4:E5:7C:B6:00:01", SENSORID_PRIMARY, .spins = 10, .x = 2.63, .y = 37.82, .z = 31.11 },
-    { "A4:E5:7C:B6:00:02", SENSORID_PRIMARY, .spins = 10, .x = -2.38, .y = -26.8, .z = -42.78 },
-    { "A4:E5:7C:B6:00:03", SENSORID_PRIMARY, .spins = 10, .x = 11, .y =  2.2, .z =  -1 },
-    { "A4:E5:7C:B6:00:04", SENSORID_PRIMARY, .spins = 10, .x = -7, .y = -53.7, .z = -57 },
-    { "A4:E5:7C:B6:00:05", SENSORID_PRIMARY, .spins = 10, .x = -10.63, .y = -8.25, .z = -18.6 },
+    { .mac = "A4:E5:7C:B6:00:01", .sensorId = SENSORID_PRIMARY, .spins = 10, .x = 2.63, .y = 37.82, .z = 31.11 },
+    { .mac = "A4:E5:7C:B6:00:02", .sensorId = SENSORID_PRIMARY, .spins = 10, .x = -2.38, .y = -26.8, .z = -42.78 },
+    { .mac = "A4:E5:7C:B6:00:03", .sensorId = SENSORID_PRIMARY, .spins = 10, .x = 11, .y =  2.2, .z =  -1 },
+    { .mac = "A4:E5:7C:B6:00:04", .sensorId = SENSORID_PRIMARY, .spins = 10, .x = -7, .y = -53.7, .z = -57 },
+    { .mac = "A4:E5:7C:B6:00:05", .sensorId = SENSORID_PRIMARY, .spins = 10, .x = -10.63, .y = -8.25, .z = -18.6 },
 };
 
 #endif


TheDevMinerTV (approved):
LGTM

TheDevMinerTV:
Also, feel free to PR the PlatformIO.ini changes in a new PR so it can be automatically checked through CI. Don't forget to also change the `platformio-tools.ini`.

Spacefish:
Ok i will add a littlebit more, as the I2C does not work yet, i will append it to this MR!

Spacefish:
B.t.w. the code in icm20948sensor.cpp does not make any sense to me:
image

ICM20948_ODRGYR and ICM20948_ODRAXL will be 1 no matter which branch is choosen..

TheDevMinerTV (approved):

Spacefish:
Ok seems to work fine, got my BMI160 sensors today: https://www.youtube.com/shorts/CuVVX6oVG7U

lib/i2cscan/i2cscan.cpp
```diff
@@ -6,7 +6,7 @@ uint8_t portArray[] = {16, 5, 4, 2, 14, 12, 13};
 uint8_t portExclude[] = {LED_PIN};
 String portMap[] = {"D0", "D1", "D2", "D4", "D5", "D6", "D7"};
 // ESP32C3 has not as many ports as the ESP32
-#elif defined(ESP32C3)
+#elif defined(ESP32C3) || defined(ESP32C6)
```
unlogisch04, line 9: Pleas check the portArray and port Exclude, i belive they have to be different from ESP32-C3

lib/i2cscan/i2cscan.cpp
```diff
@@ -113,13 +113,13 @@ namespace I2CSCAN
 
     bool hasDevOnBus(uint8_t addr) {
         byte error;
-#if ESP32C3
+#if defined(ESP32C3) || defined(ESP32C6)
```
unlogisch04, line 116: Was this neede for ESP32C6? (ESP32 does not need it as example)

platformio-tools.ini
```diff
@@ -65,6 +65,13 @@ build_flags =
   -DESP32C3
 board = esp32-c3-devkitm-1
 
+[env:BOARD_ES32C6DEVKITC1]
+platform = https://github.com/tasmota/platform-espressif32/releases/download/2024.04.14/platform-espressif32.zip
```
unlogisch04, line 69: Pls update as on main now. I belive that should work now.

platformio.ini
```diff
@@ -100,3 +100,11 @@ upload_speed = 921600
 ;  ${env.build_flags}
 ;  -DESP32C3
 ;board = lolin_c3_mini
+
+; If you want to use a ESP32C6, you can use this (experimental)
+;[env:esp32c6]
+;platform =  https://github.com/tasmota/platform-espressif32/releases/download/2024.04.14/platform-espressif32.zip
```
unlogisch04, line 106: Pls update as on main now. I belive that should work now.

unlogisch04 (changes requested):
Please update it with the current main. Also check the portArray / PortExclude. When i did try todo mine, i had seen differences but did not brother then as it was all more alpha.

unlogisch04:
> B.t.w. the code in icm20948sensor.cpp does not make any sense to me: image
> 
> ICM20948_ODRGYR and ICM20948_ODRAXL will be 1 no matter which branch is choosen..

Oddly the ESP32-S2 can not handle as much I2C traffic as the ESP8266 or the ESP32-C3 or at least when i did test it i had often Buffer overflows on ESP32-S2.

lib/i2cscan/i2cscan.cpp
Spacefish: I will try

platformio.ini
Spacefish: https://docs.platformio.org/en/latest/boards/espressif32/esp32-c6-devkitc-1.html#frameworks <- unfortunately not :( Arduino Framework is still not supported by the official platformio libs on the C6

lib/i2cscan/i2cscan.cpp
Spacefish: Added custom PortMap for ESP32C6 and included all usable pins apart from serial TX/RX and USB D+/D-

platformio-tools.ini
Spacefish: Unfortunately doesn´t work yet..
The ESP32C6 is supported in the esp32-arduino repo, but not in the platformio esp32 platform:
https://github.com/platformio/platform-espressif32/blob/develop/boards/esp32-c6-devkitc-1.json (arduino is missing from frameworks).

If i add it manually, at least my board doesn´t like it that much, i get errors like:

'''
E (120069) adc_cali: adc_cali_create_scheme_curve_fitting(79): Calibration required eFuse bits not burnt
[120001][E][esp32-hal-adc.c:322] __analogReadMilliVolts(): adc_cali_create_scheme_x failed!
'''

the tasmota repo works though...

but that might be related to me using Pre-Release Silicon which is missing the factory ADC calibration.

Spacefish:
Made most of the changes requested.

**Ports**
I added all ports of the ESP32-C6 excluding the USB pins used for serial communication and the UART TX/RX.

**I2C retries**
Removed it as it does not seem to be necessary

**Small things**
- Removed the boot delay (wasn´t required)
- Removed special case handling for one sensor which has it for the C3 as well but which did essentially nothing

**platformio ESP32-C6 arduino support**
Unfortunately the "official" platformio platform-espressif32 package still does not support any C6 board with the arduino framework.
The only two supported C6 boards in platformio are 'esp32-c6-devkitc-1' and 'esp32-c6-devkitm-1' both list only 'espidf' as framework, even on the develop branch.

Support for the C6 was added by espressif to espressif/arduino-esp32 in 3.0.0 though. So support is probably comming to platformio sooner or later in the official packages.

For now have to stick to the tasmoto platform-espressif32.

Eirenliel (approved):

PR #328, opened 2024-05-01
 fshelper: fixed ESP8266 regression caused by abstracting FS access #321

unlogisch04:
This is fixing ESP8266 calibration storage which seems to be broken since https://github.com/SlimeVR/SlimeVR-Tracker-ESP/pull/319

Use open to check if it is a directory or not as on ESP32.

As requested in comments in https://github.com/SlimeVR/SlimeVR-Tracker-ESP/pull/321

### DIFF
--- a/src/FSHelper.cpp
+++ b/src/FSHelper.cpp
@@ -37,15 +37,9 @@ bool ensureDirectory(const char* directory) {
 		}
 	}
 
-#ifdef ESP32
-	auto dir = LittleFS.open(directory);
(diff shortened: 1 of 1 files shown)


TheDevMinerTV (approved):

unlogisch04:
@l0ud your ok with that?

TheDevMinerTV:
I greatly prefer this over l0ud's solution.

l0ud (approved):
nice cleanup

PR #329, opened 2024-05-02
Remove ESP32-C3 startup delay

kounocom:
This no longer seems needed, all serial functions still work without the delay.


unlogisch04:
I did add this timeout. 
The reason for it was, that USB is able to register on the computer and allow to see the full log. ( example for WiFi Setup, and also for more easy debug ).
I don't see any reason why that should have changed.

Spacefish:
At least you have 2s to start the monitor. It helped with the C6 when i was trying it out.. Not sure if it´s 100% required, as you can still do a reset while being connected to the serial interface via USB.

kounocom:
That's kinda what I was thinking -- if the full log is needed, user can just send a reboot command via serial
And if it hangs and doesn't respond, I'm pretty sure esptool can still send a soft reset signal

m-RNA:
I don't think it's better to turn off the delay, otherwise it's not easy to connect to the serial port to see the full Log

Eirenliel (approved):

Issue #334, opened 2024-06-24
Make defines.h more powerfull


unlogisch04:
Current situation:
debug.h is included before defines,h.

I would love to be able to override the most defines in defines.h so all configuration changes (MBE,...) can be easy done within one file. 

It would also allow to add some custom fields to the Webflasher to enable/disable some options. 
An idea but not sure if it is a valid option on point of security and storage back end.

PR #337, opened 2024-08-18
add guards to some defines

unlogisch04:
Should address #334 Make defines.h more powerfull


Eirenliel:
Shouldn't you also changed globals.h to change the order?

unlogisch04:
The order should not mater, as it is a compiler define. So it is defined for every file.

unlogisch04:
Changed order so defines.h can also change the parameters of debug.h

unlogisch04:
I belive it is nolonger needed.

PR #338, opened 2024-08-19
GET.INFO Serial Command Should Returns Battery Voltage & Level (in Percentages)

ZycaR:
This update modifies the `GET INFO` serial command to include battery voltage and battery level (displayed as a percentage).

This enhancement provides detailed battery status information in the serial output and was a big help when I was troubleshooting a connection issue with the battery sense wire.

### DIFF
--- a/src/batterymonitor.h
+++ b/src/batterymonitor.h
@@ -72,6 +72,9 @@ public:
     void Setup();
     void Loop();
 
+    float getVoltage() const { return voltage; }
+    float getLevel() const { return level; }
+
 private:
     unsigned long last_battery_sample = 0;
 #if BATTERY_MONITOR == BAT_MCP3021 || BATTERY_MONITOR == BAT_INTERNAL_MCP3021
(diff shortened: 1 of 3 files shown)


src/batterymonitor.h
```diff
@@ -83,6 +83,10 @@ class BatteryMonitor
     float voltage = -1;
     float level = -1;
 
+public:
+    float getVoltage() const { return voltage; }
+    float getLevel() const { return level; }
+
```
Vyolex, line 89: Please move the public headers together.

Vyolex (changes requested):
Seems like a useful change

src/batterymonitor.h
ZycaR: done.

TheDevMinerTV (approved):
LGTM

Vyolex (approved):

ButterscotchV (approved):

PR #340, opened 2024-08-23
Magnetometer support in sfusion

furrycoding:
Added magnetometer support for sfusion sensors
Currently, only LSM6DSR with QMC5883L via sensor hub is supported (LSM6DSO might work, DSV probably won't since I didn't even look at the datasheet lol)

Sensors that support magnetometers receive a driver for it as a template argument. That driver is structured pretty much like a normal sfusion driver. Later I want to also add an option to pass it directly to `SoftFusionSensor` (for mags that are directly connected to the main I2C bus)

### TODO:
- ~~Add axis remapping~~
- Support mag driver passed directly to `SoftFusionSensor`
- Check the registers and test LSM6DSO/DSV
- Disable mag calibartion when there's no driver
- Fine-tune magnetic disturbance rejection in VQF
- Detecting errors when I2C writes to mag fail

### DIFF
--- a/src/consts.h
+++ b/src/consts.h
@@ -63,6 +63,9 @@ enum class ImuID {
 #define IMU_LSM6DSR SoftFusionLSM6DSR
 #define IMU_MPU6050_SF SoftFusionMPU6050
 
+#define IMU_LSM6DSO_QMC SoftFusionLSM6DSO_QMC
+#define IMU_LSM6DSR_QMC SoftFusionLSM6DSR_QMC
+
 #define IMU_DEV_RESERVED 250 // Reserved, should not be used in any release firmware
 
 #define BOARD_UNKNOWN 0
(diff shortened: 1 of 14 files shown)


furrycoding:
Accidentally discovered that sfusion getting merged broke BMI160 axis remapping
[Before](https://github.com/SlimeVR/SlimeVR-Tracker-ESP/blob/83b075b804f22671ef1db2cac36874f25c3e5eee/src/sensors/SensorManager.cpp#L92) - `extraParam` gets assigned to an `int` -> no downcasting
[After](https://github.com/SlimeVR/SlimeVR-Tracker-ESP/blob/fef504e3b40f6990d98ef4f0e86ae8b0b123e37d/src/sensors/SensorManager.h#L96)  - `extraParam` gets assigned to an `uint8_t` -> downcast to 8 bits, while 18 bits are needed for full axis remap
(without mag 9 bits are needed, however the upper bit is zero in the default remap, so it doesn't break a BMI that has no mag)

PR #341, opened 2024-08-26
Add a way to toggle magnetometer in runtime

### LINKED ISSUE #230: Switching magnetometer on and off at runtime
This is something that @Fishvap requested on discord, though I vaguely recall some separate discussion about doing something like this for the BNOs.

This is a cross-cutting concern, but I'm starting with an issue on the Firmware, since it doesn't matter if we add UI for it if the trackers don't support it.

The concept is that for users with trackers that have magnetometers and appropriate firmware flashed, we can eventually add a switch to the UI somewhere (per-tracker? globally?) that switches between ignoring the magnetometer or not. This would allow people to fiddle with enabling or disabling the magnetometer without having to re-flash firmware. (one potential use-case might be for someone who is travelling, their home happens to have a good magnetic environment, but wherever they are right now isn't.)

This roughly requires:
 - [ ] In the firmware, a build variant which can choose this at runtime.
   - This is more difficult for things like the mpu9250, since many users in 6dof mode use the DMP, while users in 9dof mode can't and instead use our own sensor fusion. Including both sets of code might inflate firmware size.
     - This might end up being a choice of "use DMP, or have the ability to switch" at compile time.
   - This is easier for things like the BNOs, which do the sensor fusion for us and we can choose to request magnetometer vs no magnetometer, or for IMUs where we always need to do our own sensor fusion anyway.
     - Something that I haven't checked: what fusion algorithms do we support that are compatible with which modes?
 - [ ] In the firmware protocol, a means for the server to communicate whether a magnetometer should be considered or not to the tracker
   - Do we already have a means for the server to communicate runtime settings to trackers?
 - [ ] In the GUI, the actual toggle.

ImUrX:
- Added two packets for changing "config" of a sensor, and made config id `1` be magnetometer in all IMUs (to simplify stuff)
- Implemented this toggle only for BNO currently as that's the only sensor I been using for testing
- Added a new field in the sensor info that tells you all sensor config values

Fixes #230


src/network/connection.cpp
```diff
@@ -298,6 +298,7 @@ void Connection::sendSensorInfo(Sensor& sensor) {
 	MUST(sendByte(sensor.getSensorId()));
 	MUST(sendByte(static_cast<uint8_t>(sensor.getSensorState())));
 	MUST(sendByte(static_cast<uint8_t>(sensor.getSensorType())));
+	MUST(sendByte(static_cast<uint8_t>(sensor.getMagStatus())));
```
Eirenliel, line 301: I don't like that it's sent here. We're adding flags and configs, but still expanding the sensorInfo packet. It should be sent with the rest of the config.

src/sensors/bno080sensor.cpp
```diff
@@ -217,18 +236,21 @@ void BNO080Sensor::sendData()
...
 
-#if USE_6_AXIS && BNO_USE_MAGNETOMETER_CORRECTION
-    if (newMagData)
-    {
-        newMagData = false;
-        networkConnection.sendRotationData(sensorId, &magQuaternion, DATA_TYPE_CORRECTION, magCalibrationAccuracy);
-        networkConnection.sendMagnetometerAccuracy(sensorId, magneticAccuracyEstimate);
+//    if(isMagEnabled()) {
+//        networkConnection.sendMagnetometerAccuracy(sensorId, magneticAccuracyEstimate);
+//    }
```
TheDevMinerTV, line 244: Just remove it

src/sensors/bno080sensor.cpp
```diff
@@ -113,52 +128,55 @@ void BNO080Sensor::motionLoop()
...
-    #endif // ENABLE_INSPECTION
+            uint8_t acc;
+            Vector3 nAccel;
+            imu.getLinAccel(nAccel.x, nAccel.y, nAccel.z, acc);
+            setAcceleration(nAccel);
+        }
+#endif // SEND_ACCELERATION
 
-            newMagData = true;
+#if BNO_USE_MAGNETOMETER_CORRECTION
```
Eirenliel, line 163: This can be removed.

src/sensors/bno080sensor.cpp
```diff
@@ -217,18 +236,21 @@ void BNO080Sensor::sendData()
...
-    if (newMagData)
-    {
-        newMagData = false;
-        networkConnection.sendRotationData(sensorId, &magQuaternion, DATA_TYPE_CORRECTION, magCalibrationAccuracy);
-        networkConnection.sendMagnetometerAccuracy(sensorId, magneticAccuracyEstimate);
+//    if(isMagEnabled()) {
+//        networkConnection.sendMagnetometerAccuracy(sensorId, magneticAccuracyEstimate);
+//    }
+
+    if(BNO_USE_MAGNETOMETER_CORRECTION && !isMagEnabled()) {
```
Eirenliel, line 246: BNO_USE_MAGENTOMETER_CORRECTION can be removed

TheDevMinerTV (commented):
LGTM (except for that one thing)

Eirenliel (changes requested):
I don't like that mag status is an intrinsic property of a tracker, instead of a generic flag. I think it shouldn't be sent in SensorInfo, but should be sent separately in a more generic way like a list of sensor flags. Even if it's in SensorInfo, it should be more than just this one thing, imo.

The rest looks good.

src/network/connection.cpp
ImUrX: I mean sure, but then think of the tracker initialization in the server also, do we defer it to the sensor flag list? This sensor flag list when do we send it? What are we gonna use it for aside for mag?

I feel like magnetometer fits well in the sensor info as it makes sense to have it there, but if we are gonna add a flag list I want it to make sense and used for more than just magnetometer so it's something that everyone can agree on (which will take a lot longer)

src/configuration/DeviceConfig.h
```diff
@@ -1,37 +1,35 @@
...
+	OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+	THE SOFTWARE.
 */
 
 #ifndef SLIMEVR_CONFIGURATION_DEVICECONFIG_H
 #define SLIMEVR_CONFIGURATION_DEVICECONFIG_H
 
-#include "CalibrationConfig.h"
-
 namespace SlimeVR {
```
l0ud, line 27: you can do nested namespaces like
namespace SlimeVR::Configuration {
}

src/network/connection.cpp
```diff
@@ -710,6 +728,32 @@ void Connection::update() {
...
+			uint8_t sensorId = m_Packet[12];
+			uint16_t flagId = m_Packet[13] << 8 | m_Packet[14];
+			bool newState = m_Packet[15] > 0;
+			if(sensorId == UINT8_MAX) {
+				for (auto& sensor : sensors) {
+					sensor->setFlag(flagId, newState);
+				}
+			} else {
+				auto & sensors = sensorManager.getSensors();
+				if(sensorId < sensors.size()) {
```
l0ud, line 748: should it still send acknowledge on invalid sensor id?

src/network/featureflags.h
```diff
@@ -74,7 +74,7 @@ class FirmwareFeatures {
     enum EFirmwareFeatureFlags: uint32_t {
         // EXAMPLE_FEATURE,
 		B64_WIFI_SCANNING = 1,
-
+        SENSOR_CONFIG = 2,
```
l0ud, line 77: indent

src/sensors/sensor.h
```diff
@@ -75,12 +82,18 @@ class Sensor
     bool isValid() {
         return sclPin != sdaPin;
     };
+	bool isMagEnabled() {
+		return magStatus == MagnetometerStatus::MAG_ENABLED;
```
l0ud, line 86: indent

src/sensors/bno080sensor.h
```diff
@@ -27,6 +27,8 @@
 #include "sensor.h"
 #include <BNO080.h>
 
+#define FLAG_SENSOR_BNO0XX_MAG_ENABLED 1
```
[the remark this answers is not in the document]
ImUrX: i like this idea

src/network/connection.cpp
ImUrX: what should it do instead? :thinking:

src/configuration/DeviceConfig.h
ImUrX: i dont know c++, but i will figure out what this means

ImUrX: yeah will do

src/network/connection.cpp
```diff
@@ -710,6 +728,32 @@ void Connection::update() {
...
 
 			break;
+		}
+
+		case PACKET_SET_CONFIG_FLAG: {
+			// Packet type (4) + Packet number (8) + sensor_id(1) + flag_id (2) + state (1)
+			if (len < 16) {
+				m_Logger.warn("Invalid sensor config flag packet: too short");
+				break;
+			}
```
Vyolex, line 738: Not ignoring bigger packets?

Vyolex (commented):
Just a quick review for now. It's been a while since I looked at the firmware, not really a fan of the amount of duplication that is present for the sensor configs. I'm wondering if we could consolidate them into types of configs rather than have a unique one for each IMU.

src/network/connection.cpp
ImUrX: we dont do that in any other packet, i think it makes sense to have this but shouldnt we make an issue for this so all packets have the same expected behavior on that regard?

src/network/connection.cpp
```diff
@@ -557,6 +573,8 @@ void Connection::searchForServer() {
 			m_UDP.remotePort()
 		);
 		m_Logger.traceArray("UDP packet contents: ", m_Packet, len);
+#else
+		(void)len;
```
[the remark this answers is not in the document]
ImUrX: i dont know @Eirenliel or @TheDevMinerTV seem to have pushed this

src/network/connection.cpp
l0ud: At this point, maybe just ignore like too short packet?

src/sensors/bno080sensor.h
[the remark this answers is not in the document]
ImUrX: where should i put the flags? In SensorConfig?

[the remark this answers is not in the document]
ImUrX: hi?

src/network/connection.cpp
ImUrX: made an issue for this #342, will resolve this for now

src/sensors/bno080sensor.h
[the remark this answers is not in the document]
ImUrX: :sob:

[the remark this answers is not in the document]
Eirenliel: You can pick the place yourself, no need to be blocked by people that disappeared.

src/network/connection.cpp
Eirenliel: Why do we need this information during initialization?

ImUrX: because it doesnt make sense for a initialized tracker to be missing data? i can add a new enum for it on server called "WAITING" and stuff :sob:

aside from all that, how are we gonna make this work, do we just return a bitfield?

PR #343, opened 2024-09-17
Change "PASS" to "PASSWD"

ButterscotchV:
People are confusing "PASS" with meaning that the WiFi passes something somehow, I think "PASSWD" should be more obvious, or maybe we should just say "PASSWORD"? We're not really limited on space here.


Eirenliel (approved):

Eirenliel:
Do we need any server changes for it?

ImUrX:
Nope, this is just a log when u execute WIFISCAN

PR #344, opened 2024-09-17
Add dependabot & update workflows

ButterscotchV:
GitHub Actions fails now because the versions are so old. This updates those and adds dependabot so it'll automatically update them. :3


ImUrX (approved):

Vyolex (approved):
This is great, now auto formatting checks 😃

Eirenliel (approved):

PR #345, opened 2024-09-19
BMI323 sfusion implementation

jojos38:
Working BMI323 at 200hz gyro and 200hz accel

Wasn't tested with two sensors as I don't have any but works fine on one sensor

### DIFF
--- a/src/defines.h
+++ b/src/defines.h
@@ -26,9 +26,9 @@
 // ================================================
 
 // Set parameters of IMU and board used
-#define IMU IMU_BNO085
-#define SECOND_IMU IMU
-#define BOARD BOARD_SLIMEVR
+#define IMU IMU_BMI323
+#define SECOND_IMU IMU_BMI323
+#define BOARD BOARD_CUSTOM
 #define IMU_ROTATION DEG_270
 #define SECOND_IMU_ROTATION DEG_270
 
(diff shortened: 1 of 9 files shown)


src/defines.h
```diff
@@ -26,9 +26,9 @@
 // ================================================
 
 // Set parameters of IMU and board used
-#define IMU IMU_BNO085
-#define SECOND_IMU IMU
-#define BOARD BOARD_SLIMEVR
+#define IMU IMU_BMI323
+#define SECOND_IMU IMU_BMI323
+#define BOARD BOARD_CUSTOM
```
TheDevMinerTV, line 31: Please don't change these, the firmware should always compile for the official trackers.

src/defines.h
```diff
@@ -149,6 +149,24 @@ IMU_DESC_ENTRY(IMU_BMP160, PRIMARY_IMU_ADDRESS_ONE, IMU_ROTATION, PIN_IMU_SCL, P
...
+  // Therefor, BATTERY_SHIELD_RESISTANCE is set to 0 and SHIELD_R1 and R2 are adjusted instead
+  #ifndef BATTERY_SHIELD_RESISTANCE
+    #define BATTERY_SHIELD_RESISTANCE 0
+  #endif
+  #ifndef BATTERY_SHIELD_R1 
+    #define BATTERY_SHIELD_R1 36
+  #endif
+  #ifndef BATTERY_SHIELD_R2
+    #define BATTERY_SHIELD_R2 123.5
+  #endif
```
TheDevMinerTV, line 169: these should also be removed.
```suggestion
```

TheDevMinerTV (changes requested):

jojos38:
Done, I forgot to revert these sorry

Eirenliel (approved):
The libraries say that it's your copyright, but they look like they were taken from the Bosch's sources. What's the library source?

jojos38:
> The libraries say that it's your copyright, but they look like they were taken from the Bosch's sources. What's the library source?

I rewrote an optimized the code of the library, the library was originally written without OOP, I rewritten (only the required parts) with OOP and also optimized some parts to reduce memory usage and increase speed slighty

PR #349, opened 2024-10-06
ICM-45 firmware support

gorbit99:
This PR adds support for the [ICM-45686](https://invensense.tdk.com/products/motion-tracking/6-axis/icm-45686/) IMU. To achieve the best results, it also makes changes to softfusionsensor.h to accept 32-bit sensor samples as well.


src/sensors/softfusion/drivers/icm45686.h
yenmax, line 1: You might want to update line 24 to "ICM-45686" 
＝）

Eirenliel:
Seems to be superceeded by #363

PR #353, opened 2024-10-27
Add soft sensor shutdown for BNO080

samiryhann:
Adds software shutdown routine for BNO080. Adds virtual method for software shutdown in sensor class. Updates main loop.


Eirenliel:
I think you missed around 85% of the code you reference...

samiryhann:
Sorry, this was mean for a private repo.

PR #355, opened 2024-10-30
Enforce `clang-format` on the repo

ImUrX:
Enforces formatting on the whole repo, will obviously break PRs

I also added `clang-tidy` config which is a static analyzer, it can take care of setting up a naming convention for us and force us to make modern looking C++, but its config is yet not enforced on CI


Vyolex (commented):
Finally! Don't have time to look at it today, will do tomorrow. In the meantime I did try this [config](https://github.com/Vyolex/SlimeVR-Tracker-ESP/commit/e989606aca5fc3ebd19fe9e6b0e6dc2368bf7d7f) for my own branch, might be useful.

Eirenliel (approved):

PR #356, opened 2024-10-31
Add announcement of magnetometer not supported

ImUrX:
Add second sensorconfig bit as announcement of magnetometer being supported or not, as I forgot to add such a thing for the new handshake method

### DIFF
--- a/src/sensors/bno080sensor.cpp
+++ b/src/sensors/bno080sensor.cpp
@@ -68,6 +68,8 @@ void BNO080Sensor::motionSetup()
         break;
     }
 
+	magStatus = MagnetometerStatus::MAG_NOT_SUPPORTED;
+
     if(!isMagEnabled()) {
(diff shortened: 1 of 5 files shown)


src/sensors/bno080sensor.cpp
```diff
@@ -68,6 +68,8 @@ void BNO080Sensor::motionSetup()
         break;
     }
 
+	magStatus = MagnetometerStatus::MAG_NOT_SUPPORTED;
```
Eirenliel, line 71: Why is it not supported here?

Eirenliel (approved):

src/sensors/bno080sensor.cpp
ImUrX: fixed, it was for debugging

Eirenliel (approved):

PR #357, opened 2024-11-01
BMI220 firmware support

barryblueice:
I bought some cheep BMI270 for personal built SlimeVR Case by myself from a seller from Chinese AliExpress Market, the seller claim that these sensors are BMI270:

e270dce8cdefd160e8726771a7a291f0

But when I solder the sensor on, the SlimeVR tracker detect the sensor reg is 0x26 (BMI270 is 0x24):

image

After searching from google, I found a similar post on the bosch forum: [BMI270 CHIP ID reported as 0x26 instead of 0x24](https://community.bosch-sensortec.com/mems-sensors-forum-jrmujtaw/post/bmi270-chip-id-reported-as-0x26-instead-of-0x24-Puuxy63iBXZNZEl), someone found that 0x26 is actually BMI220:

image

So I decided to try to adapt the BMI220 for the tracker. Unfortunately, I can't find any information about the source code about the BMI220. But I found the firmware successfully finally from [coreboot/chrome-ec](https://github.com/coreboot/chrome-ec/blob/main/third_party/bmi220/accelgyro_bmi220_config_tbin.h), than I try to modified the relatived source code which based on BMI270, now it can work.

VID20241101230625

**However, there may still be bugs, so use them with caution.**

The usage should be the same as BMI270/BMI160.

> Attention
> According from my friends, There is no difference between the BMI270 and BMI220 from the appearance and the sensors marking, and the sensor type can only be determined from the reg code after the sensor is soldered on the PCB.


src/consts.h
```diff
@@ -37,6 +37,7 @@ enum class ImuID {
     BMI160,
     ICM20948,
     ICM42688,
+    BMI220,
```
l0ud, line 40: Please put new sensor at the end, regardless of alphabetical order, server expects particular ID.

src/sensors/softfusion/drivers/bmi220.h
```diff
@@ -0,0 +1,419 @@
...
+            return false;
+        }
+
+        setNormalConfig(gyroSensitivity);
+
+        return true;
+    }
+
+    void motionlessCalibration(MotionlessCalibrationData &gyroSensitivity)
+    {
```
l0ud, line 309: did you check if advanced stuff like CRT really works (non-zero values returned, unique for each sensor?)

l0ud (changes requested):

src/consts.h
barryblueice: Sorry for interrupting, I will change it later

src/sensors/softfusion/drivers/bmi220.h
barryblueice: No, the CRT function not work. It returned all-zero values.
I don't have any source code or datasheet related to BMI220. So I don't know how to modified the CRT function.
According to serial console, it seems like only the CRT won't work.

image

image

Eirenliel (approved):

PR #358, opened 2024-11-03
Improve WiFi logging

ButterscotchV:
Currently our WiFi stuff is logged in different ways each time we log it, this makes it all surround the SSID in singlequotes and the "OPEN" vs "PASSWD" printout now just says the encryption type (plus it never worked on ESP8266 since "OPEN" is actually 7, not 0).

I'm absolutely certain this isn't the best way of doing this so please complain about it in the reviews/comments or whatever and I can fix it.


TheDevMinerTV (approved):
LGTM

src/serial/serialcommands.cpp
```diff
@@ -248,9 +290,9 @@ namespace SerialCommands {
 			if (scanRes >= 0) {
 				logger.info("[WSCAN] Found %d networks:", scanRes);
 				for (int i = 0; i < scanRes; i++) {
-					logger.info("[WSCAN] %d:\t%02d\t%s\t(%d)\t%s",
+					logger.info("[WSCAN] %d: %02d '%s' (%d dBm) %s",
```
ImUrX, line 293: the original reason for the usages of `\t` and no `'` or ` ` is because it simplified parsing as the only people that would be able to input tab in their SSID would be people that know what the hell they are doing, a `'` or a ` ` could easily be put on the SSID and break parsing

ButterscotchV: Switched to tabs now, the `'` can be considered to always surround the SSID and be trimmed (and/or use the length printed before)

ButterscotchV:
Tested on official trackers and it works :3
Correctly differentiates between WPA2_PSK and OPEN, before this PR it always says PASSWD

PR #362, opened 2024-11-11
Fix bounds checks in BMI270 driver in `bulk_read`

jabberrock:
Both `i` and `bytes_to_read` are `size_t`, which are `unsigned long`. `i - bytes_to_read` always underflows and should have been `bytes_to_read - i` for how many bytes are left in the buffer. But since we're dealing with unsigned values, it's safer to only do additions and comparisons.

### DIFF
--- a/src/sensors/softfusion/drivers/bmi270.h
+++ b/src/sensors/softfusion/drivers/bmi270.h
@@ -398,15 +398,18 @@ struct BMI270 {
 
 		for (uint32_t i = 0u; i < bytes_to_read;) {
 			const uint8_t header = getFromFifo<uint8_t>(i, read_buffer);
-			if ((header & Fifo::ModeMask) == Fifo::SkipFrame
-				&& (i - bytes_to_read) >= 1) {
+			if ((header & Fifo::ModeMask) == Fifo::SkipFrame) {
+				if (i + 1 > bytes_to_read) {
+					// incomplete frame, nothing left to process
+					break;
+				}
 				getFromFifo<uint8_t>(i, read_buffer);  // skip 1 byte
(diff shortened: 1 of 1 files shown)


l0ud (approved):
lgtm

src/sensors/softfusion/drivers/bmi270.h
```diff
@@ -398,15 +398,18 @@ struct BMI270 {
 
 		for (uint32_t i = 0u; i < bytes_to_read;) {
 			const uint8_t header = getFromFifo<uint8_t>(i, read_buffer);
-			if ((header & Fifo::ModeMask) == Fifo::SkipFrame
-				&& (i - bytes_to_read) >= 1) {
+			if ((header & Fifo::ModeMask) == Fifo::SkipFrame) {
+				if (i + 1 > bytes_to_read) {
```
ButterscotchV, line 402: Couldn't we just do this? Also is this even necessary considering the condition of the for loop or does `getFromFifo` progress `i` (I believe this to be the case)?
```suggestion
				if (i >= bytes_to_read) {
```

jabberrock: It's just coincidence that when you need to read 1 byte, you can use either `+ 1 > bytes_to_read` or `>= bytes_to_read`. In general, when you need to read X bytes, you can only use `+ X > bytes_to_read`. I chose this so that the "has enough bytes" check is consistent across all the cases.

The for loop is insufficient because you might end up reading past the end of the buffer.

ButterscotchV (approved):

jabberrock:
Thanks for reviewing and merging @ButterscotchV !

PR #363, opened 2024-11-12
Dynamic Sfusion

gorbit99:
This PR makes the following changes:

1. IMU Support: adds support for icm45686, icm45605, and lsm6ds3 (non tr-c version)
2. Adds nonblocking calibration
    - The tracker detects when it's at rest and calibrates itself
    - The calibration is then used the next time the tracker starts up
    - This replaces manual calibration and is set as the default
 3. Adds per sensor VQF parameters, so each IMU can be fine tuned
 4. Enables VQF's MBE feature by default, and modifies VQF to be somewhat temperature aware
 5. Adds the DELCAL serial command to only delete the IMU calibration without having to delete WIFI data
 6. Adds 32 bit sensor sample support into sfusion
 7. Overclocks the ESP, so no fifo overruns happen


platformio-tools.ini
```diff
@@ -15,26 +15,32 @@ build_unflags =
 [env:BOARD_SLIMEVR]
 platform = espressif8266 @ 4.2.1
 board = esp12e
+board_build.f_cpu = 160000000L
```
Eirenliel, line 18:
```suggestion
```
Do not overclock official slimes. We don't know how stable it is and how it changes power consumption and wi-fi characteristics.

platformio-tools.ini
```diff
@@ -15,26 +15,32 @@ build_unflags =
...
 
 [env:BOARD_SLIMEVR_DEV]
 platform = espressif8266 @ 4.2.1
 board = esp12e
+board_build.f_cpu = 160000000L
```
Eirenliel, line 23:
```suggestion
```
Same as above.

src/debug.h
```diff
@@ -38,7 +38,7 @@
 	false  // monitor accel for (triple) tap events and send them. Uses more cpu,
 		   // disable if problems. Server does nothing with value so disabled atm
 #define SEND_ACCELERATION true  // send linear acceleration to the server
-
+#define USE_NONBLOCKING_CALIBRATION
```
Eirenliel, line 41:
```suggestion
#define USE_NONBLOCKING_CALIBRATION true
```
Let's check defines to `true`, so we could set them to `false` too during compilation, for example.

Eirenliel (changes requested):
I'm not finished yet...

src/sensors/softfusion/softfusionsensor.h
```diff
@@ -71,32 +95,114 @@ class SoftFusionSensor : public Sensor {
...
+	float lastTemperatureAverage;
+	float temperatureSum = 0;
+	float temperatureTimeSum = 0;
+
+	static constexpr float temperatureAveragingSeconds = 5;
+
+	float zroChangeOverTemperature = 0;
+
+	void handleTemperatureMeasurement(float temperature, float timeStep) {
+#if USE_NONBLOCKING_CALIBRATION
```
l0ud, line 110: No C define blocks in softfusion core please :<
Should be separated as different (replaceable) calibration implementation, or at least separated in less spaghetti way :<

src/sensors/softfusion/softfusionsensor.h
```diff
@@ -23,7 +23,12 @@
 
 #pragma once
 
+#include <array>
+#include <type_traits>
+#include <utility>
+
 #include "../SensorFusionRestDetect.h"
+#include "../nonblockingcalibration/NonBlockingCalibration.h"
```
l0ud, line 31: This is subjective but I'm not the fan of "nonblocking" name for that. I think something like "auto" or "self-learning" is more descriptive for what it is.

src/sensors/nonblockingcalibration/AccelBiasCalibrationStep.h
```diff
@@ -0,0 +1,154 @@
...
+
+	THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+	IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+	FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+	AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+	LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+	OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+	THE SOFTWARE.
+*/
+
```
l0ud, line 23: Not related strictly to the code, I would like to have some description (in PR I guess) about  how that autocalibration works (both gyro and especially accel).

src/sensors/softfusion/drivers/bmi270.h
```diff
@@ -45,12 +45,30 @@ struct BMI270 {
...
 
 	static constexpr float MagTs = 1.0 / 100;
 
 	static constexpr float GyroSensitivity = 32.768f;
-	static constexpr float AccelSensitivity = 2048.0f;
+	static constexpr float AccelSensitivity = 8192.0f;
+
+	// Temperature stability constant - how many degrees of temperature for the bias to
+	// change by 0.01 Though I don't know if it should be 0.1 or 0.01, this is a guess
+	// and seems to work better than 0.1
```
l0ud, line 58: It's neither 0.01 nor 0.1 for any sensor.

l0ud (changes requested):

Eirenliel (approved):

PR #367, opened 2024-11-23
[sfusion, BMI270] Perform gryo motionless calibration before gyro offset calibration

jabberrock:
On a normal reset, we set the CRT (motionless) registers at startup, and then apply the calibrated gryo offset to every gyro sample.

Previously, we performed gyro offset calibration before CRT calibration. This means that we are actually calculating the gyro offset based on a CRT of 0, instead of the actual CRT.

This change performs CRT calibration before gyro calibration, so that the gryo offset is based on the actual CRT.

Cherry-picked from:
https://github.com/l0ud/SlimeVR-Tracker-ESP-BMI270/commit/2424b2ff08197c6b2d236714d8ca6de929c6bbe9
https://github.com/l0ud/SlimeVR-Tracker-ESP-BMI270/commit/cf349886c1e1bf562ea033df31ec04e52e83358f

### DIFF
--- a/src/sensors/softfusion/drivers/bmi270.h
+++ b/src/sensors/softfusion/drivers/bmi270.h
@@ -363,6 +363,11 @@ struct BMI270 {
 			gyroSensitivity.z = crt_values[2];
 		}
 
+		// CRT seems to leave some state behind which isn't persisted after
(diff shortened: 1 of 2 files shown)


l0ud (approved):
LGTM

ButterscotchV (approved):

PR #368, opened 2024-11-24
Somatic batch 2 changes

SomaticTom:
This will be ahead of the firmware that was shipped with devices, but should be functionally equivalent


SomaticTom:
Git can be a pain! Merge request sent to Mainline Slime instead of our repo. OOPS!

PR #371, opened 2024-12-07
Glove

Eirenliel:
Pretty complex PR that changes how defines.h works, so it needs changes from all firmware tools @ButterscotchV @loucass003 

## Changes
- Abstract hardware interface for communicating with IMUs: can now easily add support for multiple pinouts, muxers, shifters, extenders, etc. by just adding a class file
  - Interfaces currently implemented: direct pin i2c, direct int pin, PCA9546A I2C expander, MCP23X17 GPIO expander
  - Sensor defines changed to SENSOR_DESC_LIST and SENSOR_DESC_ENTRY, and type of hardware interface is required. DIRECT_WIRE for directly connected I2C, DIRECT_PIN for directly connected INT. This is the breaking change for firmware tools.
- Send sensor position information (joint the seonsor controls, not 3d position) and sensor data type as defined in flex and glove PR's of the server
  - Position is set via new defines SENSOR_INFO_LIST and SENSOR_INFO_ENTRY, see examples in defines.h
- Reports and saves BNO085 calibration data. It seems to be pretty useless, but it's a bit of information...
- Disables BNO085 accel calibration 1 minute after startup. Was hoping to fix stomping issue with it, but alas...

## Defines changes

1. Maximum sensors count
```diff
- #define MAX_IMU_COUNT 2
+ #define MAX_SENSORS_COUNT 2
```
2. IMU list is now sensor list
```diff
- #define IMU_DESC_LIST \
+ #define SENSOR_DESC_LIST \
```
```diff
- IMU_DESC_ENTRY(...)
- SENSOR_DESC_ENTRY(...)
```
3. Sensor pinouts/hardware interface are now need to be wrapped in helpers:
```diff
- IMU_DESC_ENTRY(IMU,  PRIMARY_IMU_ADDRESS_ONE, IMU_ROTATION, PIN_IMU_SCL, PIN_IMU_SDA, PRIMARY_IMU_OPTIONAL, PIN_IMU_INT) 
+ SENSOR_DESC_ENTRY(IMU, PRIMARY_IMU_ADDRESS_ONE, IMU_ROTATION, DIRECT_WIRE(PIN_IMU_SCL, PIN_IMU_SDA), PRIMARY_IMU_OPTIONAL, DIRECT_PIN(PIN_IMU_INT), 0)
```
4. Following hardware interfaces supported:
  - `DIRECT_WIRE(scl, sda)` - I2C on GPIO
  - `PCA_WIRE(scl, sda, pca_addr, pca_channel)` - I2C on PCA9546A I2C expander
  - `DIRECT_PIN(pin)` - INT pin on GPIO
  - `MCP_PIN(pin)` - INT pin on MCP23X17 GPIO extender
  - `NO_PIN` - INT pin not used
5. Added SENSOR_INFO_LIST and SENSOR_INFO_ENTRY where currently a default sensor position can be defined

## Test plan

There are a lot of changes, so we should do some testing before it's merged. I did some, but we need help with testing with other IMUs

* [x] Test with BNO085 glove + ESP32-C3 MCU + PCA9546A I2C extender + MCP23X17 GPIO extender
* [x] Test with official BNO085 slime and ESP12-F
* [ ] Test with official BNO085 slime + official BNO085 extension
* [ ] Test ESP32 with any IMU
* [ ] Test ESP32-C3 with any other IMU
* [x] Test ESP12-F with any other IMU
* [ ] Test with BMI160
* [ ] Test with BMI270
* [ ] Test with LSM6DSV
* [ ] Test with ICM-42688
* [x] Test with ICM-45686
* [ ] Test with different IMU as extension compared to main
* [ ] Test with more than 2 IMUs on different I2C pins

I'm once again asking for help with testing!

Other important stuff:
* [x] Prepare firmware tools for this change for FW 0.6.0+
* [ ] Bump firmware and protocol versions

## WHY, EIREN??
Now we can do THIS in mainline firmware!
IMG20241207135122

## Future stuff
Currently with 10 BNO sensors I get only 10 TPS on all of them, even when not all of them move. I think it can be optimized, and maybe we should start using additional cores where available. But that's out of scope right now, the goal was to support more sensors, not to make them work amazingly.


TheDevMinerTV (changes requested):
Please **rebase** to main and also move all of the unnecessary formatting commits into a different PR. It's impossible to review this PR otherwise.

src/main.cpp
```diff
@@ -65,14 +72,15 @@ void setup() {
...
-		PIN_IMU_SCL
-	);  // Make sure the bus isn't stuck when resetting ESP without powering it down
+	// Make sure the bus isn't stuck when resetting ESP without powering it down
 	// Fixes I2C issues for certain IMUs. Previously this feature was enabled for
 	// selected IMUs, now it's enabled for all. If some IMU turned out to be broken by
 	// this, check needs to be re-added.
+	auto clearResult = I2CSCAN::clearBus(PIN_IMU_SDA, PIN_IMU_SCL);
+	if (clearResult != 0) {
+		logger.error("Can't clear I2C bus, error %d", clearResult);
+		delay(60000);
```
Eirenliel, line 82: Note to self: remove delay so OTA doesn't break.

Eirenliel:
@adabru sorry for not replying earilier. The discussion is currently in private channels.

It's out of scope how many sensors you would use, this PR adds support for multiplexing to have as many sensors as you might want. The actual skeleton computation is done on the server and it supports 1 to 3 sensors per finger.

This PR is also not the place to discuss actual glove design

Eirenliel:
> Please **rebase** to main and also move all of the unnecessary formatting commits into a different PR. It's impossible to review this PR otherwise.

Sorry but this needs to be reviewed as a whole, not as a history. A lot of changes were rethought during the development. If you can't review it, you can just not, it's fine.

lib/bno080/PinInterface.h
```diff
@@ -0,0 +1,37 @@
...
+#define _H_PIN_INTERFACE_
+
+#include <Arduino.h>
+
+class PinInterface
+{
+public:
+    virtual int digitalRead();
+    virtual void pinMode(uint8_t mode);
+    virtual void digitalWrite(uint8_t val);
```
gorbit99, line 33:
```suggestion
    virtual int digitalRead() = 0;
    virtual void pinMode(uint8_t mode) = 0;
    virtual void digitalWrite(uint8_t val) = 0;
```
I don't see any direct implementation for these, they should be pure virtual then.

lib/bno080/BNO080.h
```diff
@@ -237,11 +239,13 @@ class BNO080
...
 	float getMagZ();
 	uint8_t getMagAccuracy();
-	
+
 	void endCalibration();
 	void saveCalibration();
 	void requestCalibrationStatus(); //Sends command to get status
-	boolean calibrationComplete();   //Checks ME Cal response for byte 5, R0 - Status
+	bool calibrationComplete();   //Checks ME Cal response for byte 5, R0 - Status
+	bool hasNewCliabrationStatus();
```
gorbit99, line 247: Nit: Typo

lib/bno080/PinInterface.h
```diff
@@ -0,0 +1,37 @@
...
+    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+    IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+    AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+    OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+    THE SOFTWARE.
+*/
+#ifndef _H_PIN_INTERFACE_
+#define _H_PIN_INTERFACE_
```
gorbit99, line 24: Idea: What do you think of potentially moving to #pragma once completely? I don't think I was born yet when that was fully supported by all compilers already

src/main.cpp
```diff
@@ -48,8 +49,14 @@ unsigned long loopTime = 0;
...
 bool secondImuActive = false;
 BatteryMonitor battery;
+TPSCounter tpsCounter;
 
 void setup() {
+	// For some reason, serial on C3 just doesn't work without the delay
+	// It can probably be less, but it needs investigation
+#ifdef ESP32C3
+	delay(3000);
+#endif
```
gorbit99, line 59: This was removed in a different PR (https://github.com/SlimeVR/SlimeVR-Tracker-ESP/pull/329), I assume you mean the logs get cut off at the beginning? A ring buffer should be implemented for logs to solve this imo

src/network/connection.h
```diff
@@ -84,6 +84,9 @@ class Connection {
 	// PACKET_FEATURE_FLAGS 22
 	void sendFeatureFlags();
 
+	// PACKET_FLEX_DATA 24
```
gorbit99, line 87: 24? Below it has 26 for this packet

src/sensorinterface/I2CWireSensorInterface.cpp
```diff
@@ -0,0 +1,71 @@
...
+uint8_t activeSDAPin;
+bool isI2CActive = false;
+
+namespace SlimeVR {
+void swapI2C(uint8_t sclPin, uint8_t sdaPin) {
+	if (sclPin != activeSCLPin || sdaPin != activeSDAPin || !isI2CActive) {
+		Wire.flush();
+#if ESP32
+		if (isI2CActive) {
+		} else {
```
gorbit99, line 40: Negate and remove unnecessary branch

src/sensorinterface/I2CWireSensorInterface.cpp
```diff
@@ -0,0 +1,71 @@
...
+			// Reset HWI2C to avoid being affected by I2CBUS reset
+			Wire.end();
+		}
+		// Disconnect pins from HWI2C
+		gpio_set_direction((gpio_num_t)activeSCLPin, GPIO_MODE_INPUT);
+		gpio_set_direction((gpio_num_t)activeSDAPin, GPIO_MODE_INPUT);
+
+		if (isI2CActive) {
+			i2c_set_pin(I2C_NUM_0, sdaPin, sclPin, false, false, I2C_MODE_MASTER);
```
gorbit99, line 49: Q: I'm not very familiar with this part of the code, any reason this is suddenly doing esp-idf function calls instead of the arduino equivalent?

src/sensorinterface/MCP23X17PinInterface.h
```diff
@@ -0,0 +1,64 @@
...
+#define MCP_GPB5 13
+#define MCP_GPB6 14
+#define MCP_GPB7 15
+
+/**
+ * Pin interface to use MCP23008/17 I2C GPIO port extenders
+ */
+class MCP23X17PinInterface : public PinInterface {
+public:
+	MCP23X17PinInterface(Adafruit_MCP23X17* mcp, uint8_t pin)
```
gorbit99, line 51: Definitely should be a reference

src/sensorinterface/SensorInterface.h
```diff
@@ -0,0 +1,43 @@
...
+*/
+
+#ifndef SENSORINTERFACE_H
+#define SENSORINTERFACE_H
+
+namespace SlimeVR {
+class SensorInterface {
+public:
+	virtual void init();
+	virtual void swapIn();
```
gorbit99, line 31:
```suggestion
	virtual void init() = 0;
	virtual void swapIn() = 0;
```

Should be pure virtual

src/sensors/ADCResistanceSensor.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+#include "GlobalVars.h"
+
+void ADCResistanceSensor::motionLoop() {
+#if ESP8266
+	float voltage = ((float)analogRead(m_Pin)) * ADCVoltageMax / ADCResolution;
+	m_Data = m_ResistanceDivider
+		   * (ADCVoltageMax / voltage - 1.0f);  // Convert voltage to resistance
+#endif
+#if ESP32
```
gorbit99, line 33:
```suggestion
#elif ESP32
```

Nit: combine

src/consts.h
```diff
@@ -44,6 +44,7 @@ enum class ImuID {
 	LSM6DSR,
 	ICM45686,
 	ICM45605,
+	ADC_RESISTANCE,
```
gorbit99, line 47: Prolly should rename the enum at this point

src/sensors/ADCResistanceSensor.h
```diff
@@ -0,0 +1,70 @@
...
+		float resistanceDivider,
+		float smoothFactor = 0.1f
+	)
+		: Sensor(
+			"ADCResistanceSensor",
+			ImuID::ADC_RESISTANCE,
+			id,
+			pin,
+			0.0f,
+			new SlimeVR::EmptySensorInterface
```
gorbit99, line 46: Instead of new-ing it, it would be better to have a single global instance of the empty sensor interface and use that

src/sensors/SensorManager.cpp
```diff
@@ -23,12 +23,16 @@
 
 #include "SensorManager.h"
 
+#include <map>
```
gorbit99, line 26: Map is the slower of the two map variants, while this isn't part of any hot code paths, it also doesn't take much to swap it out to an unordered_map

src/sensors/SensorManager.cpp
```diff
@@ -66,44 +70,58 @@ using SoftFusionICM45686
...
-		if (running) {
-			i2c_set_pin(I2C_NUM_0, sdaPin, sclPin, false, false, I2C_MODE_MASTER);
-		} else {
-			Wire.begin(static_cast<int>(sdaPin), static_cast<int>(sclPin), I2C_SPEED);
-			Wire.setTimeOut(150);
+void SensorManager::setup() {
+	std::map<int, DirectPinInterface*> directPinInterfaces;
+	std::map<int, MCP23X17PinInterface*> mcpPinInterfaces;
+	std::map<std::tuple<int, int>, I2CWireSensorInterface*> i2cWireInterfaces;
+	std::map<std::tuple<int, int, int, int>, I2CPCASensorInterface*> pcaWireInterfaces;
```
gorbit99, line 77: These don't need to be pointers, since they aren't stored as their base type

src/consts.h
```diff
@@ -147,6 +149,14 @@ enum class ImuID {
 #define MCU_HARITORA 8  // Used by Haritora/SlimeTora
 #define MCU_DEV_RESERVED 250  // Reserved, should not be used in any release firmware
 
+#define SENSOR_DATATYPE_ROTATION 0
+#define SENSOR_DATATYPE_FLEX_RESISTANCE 1
+#define SENSOR_DATATYPE_FLEX_ANGLE 2
+
+#define TRACKER_TYPE_SVR_ROTATION 0
+#define TRACKER_TYPE_SVR_GLOVE_LEFT 1
+#define TRACKER_TYPE_SVR_GLOVE_RIGHT 1
```
gorbit99, line 158: Should be enum classes

src/sensors/sensor.h
```diff
@@ -98,6 +101,16 @@ class Sensor {
...
 
+	virtual uint8_t getDataType() { return SENSOR_DATATYPE_ROTATION; };
+
+	uint16_t getSensorPosition() { return m_SensorPosition; };
+
+	void setSensorInfo(uint16_t sensorPosition) { m_SensorPosition = sensorPosition; };
+
+	TPSCounter m_tpsCounter;
+	TPSCounter m_dataCounter;
+	std::shared_ptr<SlimeVR::SensorInterface> m_hwInterface;
```
gorbit99, line 112: this probably should be a unique_ptr, at least I don't see it being copied anywhere

src/sensors/sensorposition.h
```diff
@@ -0,0 +1,81 @@
...
+#define POSITION_RIGHT_MIDDLE_DISTAL 44
+#define POSITION_RIGHT_RING_PROXIMAL 45
+#define POSITION_RIGHT_RING_INTERMEDIATE 46
+#define POSITION_RIGHT_RING_DISTAL 47
+#define POSITION_RIGHT_LITTLE_PROXIMAL 48
+#define POSITION_RIGHT_LITTLE_INTERMEDIATE 49
+#define POSITION_RIGHT_LITTLE_DISTAL 50
+
+#define GLOVE_LEFT 1
+#define GLOVE_RIGHT 2
```
gorbit99, line 79: Should be enum classes

src/main.cpp
Eirenliel: See discord: doesn't work for me. actually getting no serial if i don't put the delay, and i couldn't figure out why yet.

src/sensorinterface/I2CWireSensorInterface.cpp
Eirenliel: This has been written by someone else and i moved it here. I'm not sure, I can investigate indepth it if it's important...

src/sensors/ADCResistanceSensor.h
Eirenliel: doesn't matter, it happens once and it's empty. instead of bothering with scope it's easier to do that, imo.

src/sensors/SensorManager.cpp
Eirenliel: i tried and the constructor complained, so i just think it doesn't matter since it's not used in runtime

lib/bno080/BNO080.h
```diff
@@ -344,4 +349,7 @@ class BNO080
 	int16_t gyro_Q1 = 9;
 	int16_t magnetometer_Q1 = 4;
 	int16_t angular_velocity_Q1 = 10;
+
+	bool _hasNewCliabrationStatus = false;
+	uint8_t _calibrationResponseStatus, _accelCalEnabled, _gyroCalEnabled, _magCalEnabled, _planarAccelCalEnabled, _onTableCalEnabled;
```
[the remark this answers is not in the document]
Eirenliel: i really don't want to dig through bno datashet rn, it's basically useless anyway...

gorbit99 (approved):
I'll try cleaning up my remaining complaints in a follow-up PR down the line

PR #374, opened 2025-01-04
ICM45 implementation

gorbit99:
This PR implements the ICM45 family of IMUs, including the ICM-45686 and ICM-45605.

ICM-45686 supports attaching an external clock to improve the IMU sensitivity, the implementation assumes that this is done and supports it by default.

For better performance, sfusion is also changed to accept 32 bit raw sensor data depending on the IMU, which is supported by the IMUs in their so-called hi-res mode. In future commits it could be possible to bring the ICM42688 implementation up to par and implement hi-res mode (and clock) for that as well, but that is beyond the scope of this PR.


src/defines.h
```diff
@@ -26,7 +26,7 @@
 // ================================================
 
 // Set parameters of IMU and board used
-#define IMU IMU_BNO085
+#define IMU IMU_ICM45605
```
Eirenliel, line 29: Defauly should be BNO085

Eirenliel (changes requested):

src/defines.h
gorbit99: why even did that change, I hate git

Eirenliel (approved):

kounocom (approved):
Tested, works

ShineBrightMeow (approved):

PR #375, opened 2025-01-05
Dynamic Sfusion Attempt 2

gorbit99:
(Attempt 2)

This PR makes the following changes:

## IMU support

Implemented hi-res mode for icm42 (larely untested)
Added VQF params tuneable by sensor.
VQF's MBE (motion bias estimation) feature also got enabled.

## Sfusion calibration changes

Moved the original "classical" implementation into separate class called SfusionCalibration (might be a confusing name, suggestions are open).
Alongside that this PR implements "nonblocking" or "automatic" calibration, where the tracker tries to calibrate itself, whenever it detects a rest period. These new calibraions are then applied the next session.
Nonblocking calibration was made the default.


kounocom:
- [ ] Confirm that any combination of 2 sfusion IMUs works on overclocked ESP8266
- [ ] Confirm that any combination of 2 sfusion IMUs works on ESP32-C3
- [ ] Fix ICM-42688-P high-res mode
- [ ] Use the relationship between both gyro calibration points instead of hardcoded TemperatureZROChange for the TemperatureGradientCalculator whenever present
- [ ] For basic accelerometer calibration, all 3 axes can be calibrated at once
- [ ] For advanced accelerometer calibration, save the readings from all 6 sides, then once they're all present, use magneto to compute and apply the values on next boot
- [ ] Use separate rest detection instances with separate parameters for rest bias estimation and sensor calibration steps

I'll edit this to tick off completed tasks, or to add more tasks when I remember them

Siraxa:
tried this with BMI270's and aux wasn't calibrating I changed back to 
		.biasSigmaInit = 0.5f,
		.biasClip = 2.0f,
		.restThGyr = 2.0f,
		.restThAcc = 0.5f,
and it works fine, I guess Gyroscope th was a bit too small

Eirenliel:
Can we please rename it to DynamicCalibration instead of NonBlockingCalibration? Becaue the latter is not descriptive, especially in the config enum.

Eirenliel (approved):

PR #376, opened 2025-01-13
Add IMU timeout detection to `SoftFusionSensor`

pembem22:
Adds IMU timeout detection: sends an error and changes status if no data received for more than 2 seconds. Copypasted from ICM20948 code, tested with LSM6DSO.


src/sensors/softfusion/softfusionsensor.h
```diff
@@ -199,6 +199,20 @@ class SoftFusionSensor : public Sensor {
 		, m_sensor(I2CImpl(i2cAddress), m_Logger) {}
 	~SoftFusionSensor() {}
 
+	void checkSensorTimeout() {
+		uint32_t now = micros();
+		if (m_lastRotationUpdate + 2_000_000 < now) {
```
gorbit99, line 204: Flip this if around and return early instead.
Also turn 2000000 into a static constexpr variable.

src/sensors/softfusion/softfusionsensor.h
```diff
@@ -199,6 +199,20 @@ class SoftFusionSensor : public Sensor {
...
+			working = false;
+			m_status = SensorStatus::SENSOR_ERROR;
+			m_Logger.error(
+				"Sensor timeout I2C Address 0x%02x delaytime: %d ms",
+				addr,
+				now - m_lastRotationUpdate
+			);
+			networkConnection.sendSensorError(this->sensorId, 1);
```
gorbit99, line 212: Is there an enum class for these errors? If not, might be worth making one.

src/sensors/softfusion/softfusionsensor.h
gorbit99: Might be worth switching to 2e6 notation.

src/sensors/softfusion/softfusionsensor.h
pembem22: As far as I can tell, it's only used by BNO080 as it can get the error code from the sensor, so it just sends that value. ICM simply has 1, I copied that too.

src/sensors/softfusion/softfusionsensor.h
pembem22: Done

src/sensors/softfusion/softfusionsensor.h
gorbit99: High time to give it a name then, since it's used in 2 places

src/network/packets.h
```diff
@@ -67,4 +67,6 @@
 #define PACKET_INSPECTION_DATATYPE_INT 1
 #define PACKET_INSPECTION_DATATYPE_FLOAT 2
 
+#define PACKET_ERROR_SENSOR_TIMEOUT 1
```
gorbit99, line 70: Instead of a define, this should be an enum class

pembem22: Ah ok, should I put it in this file too?

gorbit99: It can probably fit in there honestly

gorbit99 (approved):
lgtm

kounocom (approved):

pembem22:
I've been getting random timeouts of both sensors every once in a while, which makes sense considering that `millis()` overflow in a bit more than one hour. I think the last commit should fix that.

gorbit99:
Not sure if it fixes it, everything is an uint, so it will wrap around when you subtract. Personally I'd switch to millis instead of micros, that has a wraparound of about 50 days, and it's not like you need the accuracy of micros.

gorbit99:
Should be good now

unlogisch04:
would it not better to fix that here <https://github.com/SlimeVR/SlimeVR-Tracker-ESP/blob/main/src/sensors/softfusion/softfusionsensor.h#L231> ?

src/network/packets.h
```diff
@@ -67,4 +67,8 @@
 #define PACKET_INSPECTION_DATATYPE_INT 1
 #define PACKET_INSPECTION_DATATYPE_FLOAT 2
 
+enum class PacketErrorCode : uint8_t {
```
unlogisch04, line 70: That seems not to be defined till now. On BNO it sends the following.
<https://github.com/SlimeVR/SlimeVR-Tracker-ESP/blob/main/src/sensors/bno080sensor.cpp#L220-L223>
//Get the reason for the last reset
//1 = POR, 2 = Internal reset, 3 = Watchdog, 4 = External reset, 5 = Other

unlogisch04 (commented):
Not sure if it does matter. Else it looks good to me.

Eirenliel (approved):

PR #378, opened 2025-01-17
Make I2Cscan non-blocking

### LINKED ISSUE #377: Flashing misconfigured firmware bricks tracker
Trackers should continue to function at least to allow OTA even if the IMUs are misconfigured, currently it just freezes and prints to serial, which means you can't use OTA to fix the issue.

kounocom:
Makes I2Cscan non-blocking, resolves #377. I could only test it with one IMU on ESP8266, where it still connected to wifi even when the IMU was not present at all, or on other pins/address. Needs further testing.

### DIFF
--- a/lib/i2cscan/i2cscan.cpp
+++ b/lib/i2cscan/i2cscan.cpp
@@ -65,10 +65,8 @@ namespace I2CSCAN
             Wire.end();
 #endif
             // Select next port
-            currentSCL++;
             if (currentSCL >= validPorts.size()) {
                 currentSCL = 0;
-                currentSDA++;
                 if (currentSDA >= validPorts.size()) {
                     scanState = ScanState::DONE;
                     return;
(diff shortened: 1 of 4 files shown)


lib/i2cscan/i2cscan.cpp
```diff
@@ -23,47 +21,79 @@ uint8_t portExclude[] = {LED_PIN};
 
 namespace I2CSCAN
 {
+    enum ScanState {
```
gorbit99, line 24: Turn this into an enum class instead

lib/i2cscan/i2cscan.cpp
```diff
@@ -23,47 +21,79 @@ uint8_t portExclude[] = {LED_PIN};
...
+        IDLE,
+        SCANNING,
+        DONE
+    };
+
+    ScanState scanState = IDLE;
+    uint8_t currentI = 0;
+    uint8_t currentJ = 0;
```
gorbit99, line 32: Would probably be best to give these meaningful names, like "currentSDA/currentSCL" or something

lib/i2cscan/i2cscan.cpp
```diff
@@ -23,47 +21,79 @@ uint8_t portExclude[] = {LED_PIN};
...
-            return addr2;
-        }
-        if (scanIfNotFound) {
-            Serial.println("[ERR] I2C: Can't find I2C device on provided addresses, scanning for all I2C devices and returning");
-            I2CSCAN::scani2cports();
-        } else {
-            Serial.println("[ERR] I2C: Can't find I2C device on provided addresses");
+    void scani2cports()
+    {
+        if (scanState == IDLE) {
```
gorbit99, line 38: Invert this if and return early

lib/i2cscan/i2cscan.h
```diff
@@ -6,6 +6,7 @@
 
 namespace I2CSCAN {
     void scani2cports();
+    void continueScan();
```
gorbit99, line 9: For consistency, I'd personally rename this to "update" or "tick"

lib/i2cscan/i2cscan.cpp
```diff
@@ -5,16 +5,14 @@
 uint8_t portArray[] = {16, 5, 4, 2, 14, 12, 13};
 uint8_t portExclude[] = {LED_PIN};
```
gorbit99, line 6: portExclude is unused

gorbit99 (approved):
lgtm

kounocom:
Tested to behave in a non-blocking way on official Slimes too when the IMU is misconfigured, though mine just prints that there's an I2C device on every address and pin, even with main firmware...

unlogisch04:
Why remove most of the comments?
Why remove port exclude? (LED or Flash might be affected)

kounocom:
portExclude is unused, comments seemed to clutter the code but if that's a dealbreaker I can add them back

unlogisch04:
I agree that it is a little mess with the comments.
But also the comments are documentation what the code does. Also removing a copyright is not the correct way i believe.

For the portExclude thing, why i did add it:
https://github.com/SlimeVR/SlimeVR-Tracker-ESP/pull/378/files#diff-3d0e70933593b8764e7020c8166d224ede8629400ba68de12ba803d92dace3e4L50
It makes sure that the flash and LED still works after running the i2c scan. (Flash and LED are normal GPIO pins in other configurations). That was the idea about the portExclude.

gorbit99 (approved):

ButterscotchV (approved):
Code looks fine, will test in a few minutes, pls wait before merge

ButterscotchV:
> Code looks fine, will test in a few minutes, pls wait before merge

Works fine, tracker had BNO fuckery but it seems like it functions as intended and I was able to recover my tracker with OTA, no USB necessary 👍

Eirenliel:
Please don't merge without me testing it tomorrow or the day after.

Eirenliel (changes requested):
Wait for me testing it in one or two days.

Eirenliel (approved):

Issue #383, opened 2025-02-03
[Feature] Tracker on a table detection


loucass003:
Have the tracker tell if the user did the imu calibration or not.
This should be enabled only on IMUs that require it like the BNO

it could simply be a boolean if it is done or not.

the check should be something like:

if the tracker accelerometer reported the gravitational constant for a few seconds then we can assume they performed the calibration

loucass003:
didnt see that it was a duplicate of https://github.com/SlimeVR/SlimeVR-Tracker-ESP/issues/381. 
Closing

PR #384, opened 2025-02-03
Report rest calibration

gorbit99:
Adds rest calibration detection to the sensors, where it's possible to fetch this information in some shape or form, and includes it in the sensor info packet.

Solves #383

### DIFF
--- a/src/sensors/sensor.h
+++ b/src/sensors/sensor.h
@@ -96,7 +96,7 @@ public:
 	const Vector3& getAcceleration() { return acceleration; };
 	const Quat& getFusedRotation() { return fusedRotation; };
 	bool hasNewDataToSend() { return newFusedRotation || newAcceleration; };
-	bool completedRestCalibration();
+	inline bool hasCompletedRestCalibration() { return restCalibrationComplete; }
 
 protected:
 	uint8_t addr = 0;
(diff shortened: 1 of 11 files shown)


src/sensors/sensor.h
```diff
@@ -96,6 +96,7 @@ class Sensor {
 	const Vector3& getAcceleration() { return acceleration; };
 	const Quat& getFusedRotation() { return fusedRotation; };
 	bool hasNewDataToSend() { return newFusedRotation || newAcceleration; };
+	bool completedRestCalibration();
```
TheDevMinerTV, line 99: If this is only supposed to be called by other subclasses, please move this to the `protected` access level.

TheDevMinerTV (changes requested):
Netcode looks fine, just a nitpick

loucass003 (approved):

src/sensors/sensor.h
gorbit99: This is actually a getter, maybe would be better to rename it to "hasCompleted..."?

TheDevMinerTV: yeah, that'd would make more sense

gorbit99: Renamed it and inlined it

TheDevMinerTV (approved):
LGTM

Eirenliel (approved):

Eirenliel:
Oh that's a nice way to do it for BNO, i likey.

Eirenliel (approved):

PR #389, opened 2025-02-15
Ignore rest calibration ack if packet is too short

### LINKED ISSUE #388: BNO trackers report non-existent extensions
Since release [v0.5.3](https://github.com/SlimeVR/SlimeVR-Tracker-ESP/releases/tag/v0.5.3), BNO trackers no longer work as expected. This was originally reported at https://github.com/SlimeVR/SlimeVR-Server/issues/1307, but is firmware specfic. I will test further to provide more information.

ButterscotchV:
Sensor rest calibration ack needs to support old server versions, this PR checks the packet length first, acks by default if not returned, and fixes the packet ack byte index. Also C++ does not initialize the boolean to false, so this sets it to false by default.

Fixes #388.


Eirenliel (approved):

gorbit99 (approved):
lgtm

PR #392, opened 2025-02-22
Revert "Don't copy memory on ICM45 reads"

gorbit99:
Seems like the "Don't copy memory on ICM45 reads" commit causes reproducible and constant crashes on ESP32-C3. SInce it's a fairly small performance improvement, I'm simply reverting it to fix the issue.

### DIFF
--- a/src/sensors/softfusion/drivers/icm45base.h
+++ b/src/sensors/softfusion/drivers/icm45base.h
@@ -53,9 +53,6 @@ struct ICM45Base {
 
 	static constexpr bool Uses32BitSensorData = true;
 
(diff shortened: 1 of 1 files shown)


kounocom (approved):

kounocom:
Can repro the crashing on a separate ESP32-C3 tracker with ICM-45686

ShineBrightMeow (approved):

Eirenliel (approved):

Eirenliel (commented):
Sorry I will need to test if we still have enough performance on official trackers before it cna be merged.

PR #394, opened 2025-02-26
Re-order SensorInfo packet

ButterscotchV:
This PR is to go with https://github.com/SlimeVR/SlimeVR-Server/pull/1317

From firmware v0.5.3 to v0.5.4, the SensorInfo packet has the rest calibration byte before the default position and data type, we should probably retain compatibility with the already released firmware and follow that packet order, otherwise we will be introducing some version-specific issues.


Eirenliel (approved):

Eirenliel:
Oops, merge conflict

PR #395, opened 2025-02-27
Sfusion mag support

gorbit99:
This PR adds sensor AUX interface mag support to sfusion IMUs.

Currently the effort is put in for the following IMUs:
- LSM6DSR
- ICM45686 (currently WIP)

And the following magnetometers attached to them:
- IST8306

As these are the ones easily available/used in the wild, but the PR aims to make adding support for new IMUs or magnetometers fairly straightforward.


Eirenliel:
Instead of using a define, can you use the config boolean like in BNO? So it can be controlled from the server

gorbit99:
@Eirenliel That's indeed the idea down the line (as in before I make this PR be not a draft), but for devving, this was quicker

Eirenliel:
Also needs support for the QMC that will be on the v1.2 slimes

PR #398, opened 2025-03-08
fix IMU Address missing in VS-Studio

unlogisch04:
On Webflasher it does not build with swapped IMU-Address. On Visual Studio Code it is no longer easy to define IMU without changing manual addresses. This should fix both issues, and still allow the new flasher to work

Problem introduced by PR #372

### DIFF
--- a/src/sensors/SensorManager.cpp
+++ b/src/sensors/SensorManager.cpp
@@ -25,7 +25,7 @@
 
 #include <map>
 #include <type_traits>
-
+#include "sensoraddresses.h"
 #include "bmi160sensor.h"
 #include "bno055sensor.h"
 #include "bno080sensor.h"
--- a/src/sensors/SensorManager.cpp
+++ b/src/sensors/SensorManager.cpp
@@ -153,10 +153,8 @@ void SensorManager::setup() {
 	// Apply descriptor list and expand to entries
 	SENSOR_DESC_LIST;
 
-#define SENSOR_INFO_ENTRY(ImuID, ...)                        \
-	{                                                        \
-		m_Sensors[SensorTypeID]->setSensorInfo(__VA_ARGS__); \
-	}
+#define SENSOR_INFO_ENTRY(ImuID, ...) \
+	{ m_Sensors[SensorTypeID]->setSensorInfo(__VA_ARGS__); }
 	SENSOR_INFO_LIST;
 
 #undef SENSOR_DESC_ENTRY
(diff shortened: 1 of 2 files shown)


unlogisch04:
Seems I missed something will need to check later.
IMU gets not the right address. On phone I can't mark it as draft.

Eirenliel:
How about this fix, @unlogisch04 ? https://github.com/SlimeVR/SlimeVR-Tracker-ESP/tree/swapped-imu-addr-fix

unlogisch04:
> How about this fix, @unlogisch04 ? https://github.com/SlimeVR/SlimeVR-Tracker-ESP/tree/swapped-imu-addr-fix

Seems quite to be a complex fix.

I think for most thing its fine. But not sure if we need to add defines for:
SECONDARY_IMU_ADDRESS_ONE
PRIMARY_IMU_ADDRESS_TWO

I think we could get away without them if we can change all the Webflasher.
As i belive it should not mater if we use
PRIMARY_IMU_ADDRESS_ONE for SECONDARY_IMU_ADDRESS_ONE
SENCONDARY_IMU_ADDRESS_TWO for PRIMARY_IMU_ADDRESS_TWO
from build stuff @ShineBrightMeow @ButterscotchV 

grafik


I did test it on my webflasher for: 
BMI160
BNO085
ICM20948
MPU6500
LSM6DSO
ICM42688
I did not test older versions of the fw

unlogisch04:
Additional:
For what a class? Would not a simple function do it?

Also in IMU side. There is no indication how many addresses a IMU has. Address as a uint8_t array with all possible adresses?
I'm not sure if it is useful in the future.

Eirenliel:
A class so we can have cast from int in case we have explicit address.

I don't like yor implementation because we already have each IMU class have its address, and we should use it, instead of doubling it.

I got really confused about SECONDARY_IMU_ADDRESS_ONE and PRIMARY_IMU_ADDRESS_TWO... i can add the defines for them, i'm just not sure what they should be?

unlogisch04:
I belive this should be it:
Not sure if it was because you can have different imu on the i2c line. (not that i would recommend it)

```cpp
#ifndef SECONDARY_IMU_ADDRESS_ONE
#define SECONDARY_IMU_ADDRESS_ONE true
#endif

#ifndef PRIMARY_IMU_ADDRESS_TWO
#define PRIMARY_IMU_ADDRESS_TWO false
#endif
```

Eirenliel:
See https://github.com/SlimeVR/SlimeVR-Tracker-ESP/pull/401

PR #399, opened 2025-03-09
ICM45*: Process accel sample LSB correctly

kounocom:
we were using the 4 lowest gyro bits for the LSB bits of the accel, instead of the 4 lowest accel bits lol oopsie


gorbit99:
Can confirm, this is now correct

gorbit99 (approved):

Issue #380, opened 2025-01-30
Send error information to the server


Eirenliel:
Right now when trackers error out they don't send any information to the server about the error. Send this info in some way (even just a text is fine, but also an error code for some common errors) so it can be displayed to the user.

unlogisch04:
Can we define more exactly what is needed in addition?
The following i found, that could be the main issue for this issue?
  - BNO08x seems to be implemented wrong if you disconnect the IMU. it will not set "SensorStatus::SENSOR_ERROR" but "SensorStatus::SENSOR_OFFLINE"


What errors are missing and would help to know in server? 
- [ ] IMU not found
  - At startup, if a IMU is not found, only for optional false IMUs 
  - Currently SensorStatus SENSOR_ERROR
- [ ] IMU not found but I2C scan found (not sure if we want to go into possible new subcategory)
  - At startup, or delayed (i2c scan not blocking now) would be a new error or status message
  - Currently:
    - SensorStatus SENSOR_ERROR (maybe broken on BNO?)
  - [ ] Devices found on the same i2c lines (1-10 results more would indicate a wiring/other issue)
  - [ ] Devices found on other i2c lines (1-10 results more would indicate a wiring/other issue)
  - [ ] Wiring issue
- [ ] Connection to IMU lost
  - Most likely on extension when unplugged while running
  - Sensor behave differently on that.
  - sendSensorError on some of the sensors (BNO085, ICM20948, softFusion) (only once) (WATCHDOG_TIMEOUT on ICM20948, softFusion)
  - BNO08x seems to be implemented wrong if you disconnect the IMU. it will not set "SensorStatus::SENSOR_ERROR"
- [ ] From IMU Send Error 
  - Currently only used for BNO085, ICM20948, softFusion (only send once when the error occurs. no Buffering)
  - PacketErrorCode

Currently we have from UDP Tracker:

1. SensorStatus:
-Implemented for all IMU 
-send on connect and on change
<https://github.com/SlimeVR/SlimeVR-Tracker-ESP/blob/main/src/sensors/sensor.h#L44-L48>
```cpp
enum class SensorStatus : uint8_t {
	SENSOR_OFFLINE = 0,
	SENSOR_OK = 1,
	SENSOR_ERROR = 2
};
```

2. PacketErrorCode:
-Only used on BNO08x and ICM20948
-Send only when sensor gives a error once not on reconnect,..
[...]

PR #400, opened 2025-03-12
Send error information to the server

LunaroakF:
[SlimeVR-Tracker-ESP Issues#380](https://github.com/SlimeVR/SlimeVR-Tracker-ESP/issues/380)
Using PacketErrorCode in the packet.h [packets.h Line 72](https://github.com/SlimeVR/SlimeVR-Tracker-ESP/blob/main/src/network/packets.h)
I moved PacketErrorCode to Sensor.h in this.
- using PACKET_ERROR 14
- if I2CSCAN found an available IMU but not in the defines.h, Tracker will send PacketErrorCode::NOT_APPLICABLE to the server. else PacketErrorCode::OTHER

Server: [Receive ErrorData From Tracker And replace "Error" #1344](https://github.com/SlimeVR/SlimeVR-Server/pull/1344)
SolarXR-Protocol: [Error imformation show to the user #163](https://github.com/SlimeVR/SolarXR-Protocol/pull/163)

### DIFF
--- a/src/network/packets.h
+++ b/src/network/packets.h
@@ -24,49 +24,62 @@
 #ifndef SLIMEVR_PACKETS_H_
 #define SLIMEVR_PACKETS_H_
 
-#define PACKET_HEARTBEAT 0
-// #define PACKET_ROTATION 1 // Deprecated
-// #define PACKET_GYRO 2 // Deprecated
-#define PACKET_HANDSHAKE 3
-#define PACKET_ACCEL 4
-// #define PACKET_MAG 5 // Deprecated
-// #define PACKET_RAW_CALIBRATION_DATA 6 // Deprecated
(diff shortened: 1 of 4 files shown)


src/network/packets.h
```diff
@@ -68,14 +68,4 @@
 #define PACKET_INSPECTION_DATATYPE_INT 1
 #define PACKET_INSPECTION_DATATYPE_FLOAT 2
 
-// From the SH-2 interface that BNO08x use.
-enum class PacketErrorCode : uint8_t {
```
Eirenliel, line 72: Why moved it?

Eirenliel (changes requested):
Also needs formatting

src/network/packets.h
LunaroakF: I don't get why this in packet.h 
maybe it belongs to sensor state so I move it into sensors.h

PR #403, opened 2025-03-14
ICM45*: Remove erroneous sizeof() from FIFO read logic

kounocom:


gorbit99 (approved):
Can confirm it's correct now

Eirenliel (approved):

PR #404, opened 2025-03-16
Send NOT_APPLICABLE Error to the server

LunaroakF:
[SlimeVR-Tracker-ESP Issues#380](https://github.com/SlimeVR/SlimeVR-Tracker-ESP/issues/380)
Using PacketErrorCode in the packet.h [packets.h](https://github.com/SlimeVR/SlimeVR-Tracker-ESP/blob/main/src/network/packets.h)

- using PACKET_ERROR 14
- if I2CSCAN found an available IMU but not in the defines.h, Tracker will send PacketErrorCode::NOT_APPLICABLE to the server.

Server: [Receive ErrorData From Tracker And replace "Error" #1344](https://github.com/SlimeVR/SlimeVR-Server/pull/1344)
SolarXR-Protocol: [Error imformation show to the user #163](https://github.com/SlimeVR/SolarXR-Protocol/pull/163)

### DIFF
--- a/src/network/connection.cpp
+++ b/src/network/connection.cpp
@@ -492,7 +492,8 @@ void Connection::updateSensorState(std::vector<std::unique_ptr<Sensor>>& sensors
 	for (int i = 0; i < (int)sensors.size(); i++) {
 		if (isSensorStateUpdated(i, sensors[i])) {
 			sendSensorInfo(*sensors[i]);
-			if (I2CSCAN::isFound()) {
+			if (I2CSCAN::isFound()
+				&& sensors[i]->getSensorState() == SensorStatus::SENSOR_ERROR) {
 				sendSensorError(
 					static_cast<uint8_t>(sensors[i]->getSensorId()),
 					static_cast<uint8_t>(PacketErrorCode::NOT_APPLICABLE)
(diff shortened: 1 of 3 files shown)


Eirenliel (approved):

src/network/connection.cpp
```diff
@@ -492,6 +492,13 @@ void Connection::updateSensorState(std::vector<std::unique_ptr<Sensor>>& sensors
 	for (int i = 0; i < (int)sensors.size(); i++) {
 		if (isSensorStateUpdated(i, sensors[i])) {
 			sendSensorInfo(*sensors[i]);
+			if (I2CSCAN::isFound()
```
unlogisch04, line 495: isSensorStateUpdated is only true when some fields changing. Did you check that it works as expected?

bool Connection::isSensorStateUpdated(int i, std::unique_ptr<Sensor>& sensor) {
	return m_AckedSensorState[i] != sensor->getSensorState()
		|| m_AckedSensorCalibration[i] != sensor->hasCompletedRestCalibration()
		|| m_AckedSensorConfigData[i] != sensor->getSensorConfigData();
}

LunaroakF: it works as expected, because this error usually be detected when the tracker first connect the server. and this time sensor state should be changed.

unlogisch04: Not sure if we make the protocol a little hard.
We have the status packet that we send regularly out. 
But then we have the SensorError packet that is only send when we receive a error from the sensor.  (till now)
I think this not the best practice. What happens when disconnecting/connecting to the server, or restarting the server?
Will then the error still be shown correctly?

Do we need to send PacketErrorCode::NOT_APPLICABLE when a i2c is found?

I tested with BNO085+ESP8266:
- at startup without sensor connected -> shows error
- sensor connected but disconnected while running -> does not show error.  (maybe like before)

Currently i have mixed feelings. Do we rly improve things?

PR #406, opened 2025-03-17
add handling for not given pin (-1 or 255)

unlogisch04:
This commit fixes the use case BNO08x whit out a interrupt pin.

This fix reflects the behavior till now where <uint8_t>255 was no int pin.


Eirenliel (approved):

Eirenliel:
Good catch

PR #409, opened 2025-03-21
Initial ADS111x implementation

gorbit99:
Disclaimer: I don't currently have this IC in my hand, and the lack of flex sensor implementation makes it theoretically impossible (or at least very hard) to test. As such this will stay a draft until it's actually working with a flex sensor.

This PR implements the ADS1114 and ADS1115 I2C analogue multiplexers. This is done through the recently created PinInterface virtual class. As a related change, it also implements the analogRead method on the PinInterface and rewrites the existing ADCResistanceSensor to use that instead of a regular pin.

In theory the ADS1113 could also be supported, but the lack of a drdy pin would mean that the config register would need to be polled constantly, which the author of this pull request found particularly ugly.

### DIFF
--- a/lib/bno080/BNO080.h
+++ b/lib/bno080/BNO080.h
@@ -50,7 +50,7 @@
 #include <Wire.h>
 #include <SPI.h>
 #include <memory>
-#include "PinInterface.h"
+#include <PinInterface.h>
 
 //The default I2C address for the BNO080 on the SparkX breakout is 0x4B. 0x4A is also possible.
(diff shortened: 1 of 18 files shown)


Eirenliel (changes requested):

Eirenliel (changes requested):
Needs to be rebased

gorbit99:
Converting this to a draft, so I can implement parallel muxes first

gorbit99:
I'll update this after other PRs are merged

PR #416, opened 2025-04-06
More ICM45 fixes

kounocom:
This PR fixes 2 more issues with ICM45*: 

- SensorManager.h now scans for an IMU twice, since ICM45* [may not respond on the first I2C transaction](https://github.com/tdk-invn-oss/motion.mcu.icm45686.driver/blob/fb4275febca58596774cee0dc0654cdc058a640c/icm45686/imu/inv_imu_driver.c#L66C5-L69). No other IMUs have this issue, so the 2nd scan will never run and therefore have any effect on them.
- ICM45* will never fully empty the FIFO, instead leaving one packet at all times. Not sure why, but fully emptying it will eventually cause FIFO corruption. This might be documented in AN-000364, just like the first issue, but no one's found this document anywhere on the web yet.

### DIFF
--- a/src/sensors/SensorManager.h
+++ b/src/sensors/SensorManager.h
@@ -114,6 +114,9 @@ private:
 		if (I2CSCAN::hasDevOnBus(i2cAddress)) {
 			m_Logger
 				.trace("Sensor %d found at address 0x%02X", sensorID + 1, i2cAddress);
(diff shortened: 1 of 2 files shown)


Eirenliel:
The https://github.com/SlimeVR/SlimeVR-Tracker-ESP/tree/spi-support2 branch will be using a different way to check if the IMU is on the bus by trying to ask it directly, and also using I2C scan... so uh... might need to update it there, not here

Eirenliel (approved):

PR #419, opened 2025-04-16
Remove OTA timeout

kounocom:
umu

### DIFF
--- a/lib/ota/ota.cpp
+++ b/lib/ota/ota.cpp
@@ -64,12 +64,6 @@ void OTA::otaSetup(const char * const otaPassword) {
 
 void OTA::otaUpdate() {
     if(enabled) {
-        if(bootTime + 60000 < millis()) {
-            // Disable OTA 60 seconds after boot as protection measure
-            enabled = false;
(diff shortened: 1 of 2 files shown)


unlogisch04:
The reason for disabling the OTA update after a while is to protect the tracker from false updates.

kounocom:
> The reason for disabling the OTA update after a while is to protect the tracker from false updates.

This was requested by people at the slime cave

gorbit99:
lgtm

unlogisch04:
> > The reason for disabling the OTA update after a while is to protect the tracker from false updates.
> 
> This was requested by people at the slime cave

I don't care who requested it. It is just not safe to leave it open.
We would need to implement Signed Updates to have a secure update path. That would involve Keyhandling,...  . Currently you only have to know a public accessible password for it.

unlogisch04 (changes requested):
Can you make this as a #define for user that want a different behavior.

unlogisch04 (approved):

Eirenliel (approved):

PR #420, opened 2025-04-16
update actions.yml ubuntu

unlogisch04:


Eirenliel:
Oops merged #421 instead of this.

PR #423, opened 2025-04-18
[ICM45*] Fix processing bad samples and stack overflow panics

jabberrock:
1. At startup, we were receiving invalid samples from the IMU and passing it to VQF. This causes huge initial rotations and accelerations that takes time for VQF to eliminate.

   Fix is to check the header to see if the data is present. Also had to add a check for invalid gyro samples even though it is present.

2. During play, the tracker would rarely go haywire and jump into a random position. I suspect this is caused by a stack overflow, which causes the tracker to panic. Combined with problem (1), whenever the tracker restarts, it would send huge rotations and accelerations to the server. This causes the jump.

   Fix is to make the read_buffer static, so that it's reserved on the BSS segment instead of the stack. I noticed this because I was getting panics when I tried to increase the size of the read buffer, or declared some new stack variables.

After implementing these two fixes, rotation and acceleration are stable at startup, and I am able to make modifications to the icm45base code without randomly encountering panics.

### DIFF
--- a/src/sensors/softfusion/drivers/icm45base.h
+++ b/src/sensors/softfusion/drivers/icm45base.h
@@ -141,31 +141,41 @@ struct ICM45Base {
 		GyroCall&& processGyroSample,
 		TempCall&& processTemperatureSample
 	) {
-		const auto fifo_packets = i2c.readReg16(BaseRegs::FifoCount);
-		const auto fifo_bytes = (fifo_packets - 1) * FullFifoEntrySize;
+		// Allocate statically so that it does not take up stack space, which
+		// can result in stack overflow and panic
+		constexpr size_t MaxReadings = 8;
+		static std::array<uint8_t, FullFifoEntrySize * MaxReadings> read_buffer;
 
-		std::array<uint8_t, FullFifoEntrySize * 8> read_buffer;  // max 8 readings
-		const auto bytes_to_read = std::min(
-									   static_cast<size_t>(read_buffer.size()),
-									   static_cast<size_t>(fifo_bytes)
-								   )
-								 / FullFifoEntrySize * FullFifoEntrySize;
+		constexpr int16_t InvalidReading = -32768;
+
+		size_t fifo_packets = i2c.readReg16(BaseRegs::FifoCount);
+		fifo_packets = std::min(fifo_packets, MaxReadings);
+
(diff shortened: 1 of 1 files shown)


jabberrock:
@kounocom @gorbit99 FYI

src/sensors/softfusion/drivers/icm45base.h
```diff
@@ -141,31 +141,41 @@ struct ICM45Base {
...
 	) {
-		const auto fifo_packets = i2c.readReg16(BaseRegs::FifoCount);
-		const auto fifo_bytes = (fifo_packets - 1) * FullFifoEntrySize;
-
-		std::array<uint8_t, FullFifoEntrySize * 8> read_buffer;  // max 8 readings
-		const auto bytes_to_read = std::min(
-									   static_cast<size_t>(read_buffer.size()),
-									   static_cast<size_t>(fifo_bytes)
-								   )
-								 / FullFifoEntrySize * FullFifoEntrySize;
```
jabberrock, line 152: I also simplified this section and got rid of the `-1`, which didn't seem to be necessary. Let me know if you prefer I revert this.

Eirenliel (approved):

src/sensors/softfusion/drivers/icm45base.h
```diff
@@ -141,31 +141,41 @@ struct ICM45Base {
...
+		// can result in stack overflow and panic
+		constexpr size_t MaxReadings = 8;
+		static std::array<uint8_t, FullFifoEntrySize * MaxReadings> read_buffer;
+
+		constexpr int16_t InvalidReading = -32768;
+
+		size_t fifo_packets = i2c.readReg16(BaseRegs::FifoCount);
+		fifo_packets = std::min(fifo_packets, MaxReadings);
+
+		size_t bytes_to_read = fifo_packets * FullFifoEntrySize;
```
gorbit99, line 154: The - 1 is necessary due to an issue with the icm45, please put it back

gorbit99 (changes requested):

src/sensors/softfusion/drivers/icm45base.h
jabberrock: I brought it back, added a commend and made sure it never underflows.

kounocom (approved):
lgtm

Eirenliel (approved):

PR #424, opened 2025-04-23
Softfusion cleanup

gorbit99:
This PR cleans up some code from softfusionsensor.h and the surrounding classes

The three main differences are:
- A new IMUConsts struct that handles most of the constexpr stuff instead of having to have that in the sfusion class. This also reduces code duplication.
- A new DriverCallbacks struct, that stores the callbacks for the bulkRead method on the individual drivers. This should make it easier to extend this struct in the future and it gets rid of the templates on bulkRead
- It moves some of the leftover calibration methods into SoftFusionCalibration, since that's where they were used only


Eirenliel:
Can we maybe first merge the SPI support?

gorbit99:
Only merge this after spi-support has been merged!

kounocom (approved):

Eirenliel:
uwu please rebase

PR #431, opened 2025-04-28
Reduce Temperature Pooling for ICM45 series IMUs

nightshade231:
This change is to free up CPU time on EPS82xx series development boards for DIY trackers using dual ICM45 series IMUs.
The Accelerometer was previously reduced, however I found that the temp reporting was almost instantaneous when it didn't need to be, and also was partly to blame for trackers spinning when violently shaken, or other somewhat vigorous motions (including tapping).

### DIFF
--- a/src/sensors/softfusion/drivers/icm45base.h
+++ b/src/sensors/softfusion/drivers/icm45base.h
@@ -39,7 +39,7 @@ struct ICM45Base {
 
 	static constexpr float GyrTs = 1.0 / 409.6;
 	static constexpr float AccTs = 1.0 / 102.4;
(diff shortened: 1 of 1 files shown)


Eirenliel:
This doesn't actually change polling rate, it's just interpolation constant. Temperature is always reported by ICM.

For two imus you should use spi or just not use yep imus.

nightshade231:
> This doesn't actually change polling rate, it's just interpolation constant. Temperature is always reported by ICM.
> 
> For two imus you should use spi or just not use yep imus.

Already found this out (and the fix PR) from gorbit on the discord. I was not aware thatSPI was also added in with ESP so I'll be trying that here soon.

Issue #422, opened 2025-04-17
Move the logic of this check to the sensor class


unlogisch04:
The sensor class should tell if the config has been changed.
It would avoid the need for check
```cpp
		&& sensor->getSensorType() != SensorTypeID::Unknown
		&& sensor->getSensorType() != SensorTypeID::Empty;
```
The challenge is probably to get the acknowledged back to the sensor. 

https://github.com/SlimeVR/SlimeVR-Tracker-ESP/blob/ee48341c303e97a2d35eb978311bb682dc1cce9b/src/network/connection.cpp#L516-L522

PR #432, opened 2025-04-28
Move state update check logic to the sensor class

gorbit99:
Completes #422

### DIFF
--- a/src/network/connection.cpp
+++ b/src/network/connection.cpp
@@ -514,11 +514,11 @@ void Connection::maybeRequestFeatureFlags() {
 }
 
 bool Connection::isSensorStateUpdated(int i, std::unique_ptr<Sensor>& sensor) {
-	return (m_AckedSensorState[i] != sensor->getSensorState()
-			|| m_AckedSensorCalibration[i] != sensor->hasCompletedRestCalibration()
-			|| m_AckedSensorConfigData[i] != sensor->getSensorConfigData())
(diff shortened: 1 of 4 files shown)


unlogisch04:
Technically complete.
I was also hoping to get ride of the following variables too.
m_AckedSensorState
m_AckedSensorCalibration
m_AckedSensorConfigData

src/network/connection.cpp
```diff
@@ -691,15 +665,19 @@ void Connection::update() {
...
+							sensors[i]->getSensorConfigData()
+						);
 						break;
 					}
-					m_AckedSensorCalibration[i]
-						= sensorInfoPacket.hasCompletedRestCalibration;
+					sensors[i]->signalAckedStateUpdate(
+						sensorInfoPacket.sensorState,
+						sensorInfoPacket.hasCompletedRestCalibration,
+						sensorInfoPacket.sensorConfigData
```
unlogisch04, line 679: It seems like this line has been forgotten in the code before.
But i'm leads to that what we want. How does the server request a change as example for the magnetometer?
I think i need to test that.

unlogisch04 (commented):
Generally what i was thinking of thanks.
The only open thing to approve is the test for that line.

src/network/connection.cpp
unlogisch04: testet

unlogisch04 (approved):

PR #435, opened 2025-05-03
Rework WiFi code

gorbit99:
This PR reworks the WiFi code in the following ways:

- WiFiNetwork and WiFiProvisioning are now singleton classes, instead of namespaces with free-floating functions in them
- The code for the wifi connection logic is now pulled into separate functions for clarity
- The mcu only tries switching to the G phy mode if the connection on the N phy mode timed out, not when the credentials were incorrect
- The reason for the failure is now reported correctly


gorbit99:
Since the ESP8266 and the ESP32 code didn't agree on the status codes, I remapped them to the following error codes instead:
0 - Timeout (likely DHCP failure)
1 - SSID not found
2 - Wrong password
3 - Unknown error (didn't manage to trigger this)

The new format for the message that the connection failed is
Can't connect from any credentials, error: <error code>, reason: <error message>.

ButterscotchV (approved):
Looks good, the code looks so much better. This requires thorough testing to ensure it matches the previous behaviour. If anyone tests, I would appreciate seeing what's been tested as comments on this PR :3

Eirenliel (approved):

Issue #71, opened 2022-01-13
Fancy wifi provisioning


Eirenliel:
This is both for firmware and software. Here are features that needs to be implemented:

- [ ] Tracker creates an AP with configuration protal, if it can't connect to any wifi network (pr #59 already implements this)
- [ ] Tracker can be commanded to provision nearby trackers. With this command it tries to connect to all unconfigured trackers (it gudges by an AP name) and sends them credentials of its own wifi network
  - [ ] Server should be able to command trackers to provision other trackers remotely or via USB
- [ ] When tracker is connected via USB, but it doesn't have a wifi connection, a popup offering wifi configuration should pop up
  - [ ] This pop up should let user select out of found wifi networks, it should have an checkbox "Connect other trackers", see above

Eirenliel:
![](https://github.trello.services/images/mini-trello-icon.png) [Fancy Wi-Fi Provisioning](https://trello.com/c/Vc2GPIcA/12-fancy-wi-fi-provisioning)

loucass003:
``` When tracker is connected via USB, but it doesn't have a wifi connection, a popup offering wifi configuration should pop up```

Is this still a thing with the wifi provisioning in the gui. Because those two approches would get in conflict. 

if you are provisioning new trackers you dont want the wifi popup to show up on the wifi page of the new gui.

Eirenliel:
Yeah, that shouldn't be on the wifi page, that should be when you're not in the wizard.

electromuis:
Just added some support for wifi scanning in the gui. Might be relevant here: https://github.com/SlimeVR/SlimeVR-Server/pull/1033

PR #436, opened 2025-05-05
Fancy wifi provisioning

gorbit99:
Depends on #435  

Attempts to implement #71

This PR reworks the wifi provisioning, which originally used SmartConfig, to instead use an ESP-Now based implemenation.

When a tracker can't connect to any of its credentials options, it starts sending out esp-now packets on all possible channels. If another tracker that's currently provisioning gets these packets, it will reply with the SSID and password of the wifi network it is connected to. The searching tracker then receive these and attempt to connect with the provided credentials.

A tracker can be put into provisioning mode by either sending it the "START PROVISION" serial command, or sending it a WifiProvisioning packet (packet id 27, only parameter a boolean saying if the provisioning should be started or stopped).

Caveats:
- Seems like the range on the esp-now packets - at least using official slime boards - isn't great. The tracker needs to be basically next to the provisioning one to catch the packets. From reading online, this might be somewhat determined by the 2.4ghz channel we use for sending (which needs to be the same as the channel of the WiFi network), and the attenuation might also play a role. I propose to call this a security feature.

TODO:
- [ ] Finetune timeout and interval parameters. Increasing the latter *might* make catching packets a bit more consistent
- [ ] Test extensively, both on ESP8266 and ESP32

### DIFF
--- a/src/defines.h
+++ b/src/defines.h
@@ -28,7 +28,7 @@
 // Set parameters of IMU and board used
 #define IMU IMU_AUTO
 #define SECOND_IMU IMU_AUTO
-#define BOARD BOARD_SLIMEVR_V1_2
+#define BOARD BOARD_SLIMEVR
 #define IMU_ROTATION DEG_270
 #define SECOND_IMU_ROTATION DEG_270
 
--- a/platformio.ini
+++ b/platformio.ini
@@ -52,7 +52,7 @@ build_flags =
 
 build_unflags = -Os -std=gnu++11 -std=gnu++17
 
-; If you want to enable OTA Updates, uncomment and set OTA password here and in credentials.h
+; If you want to enable OTA Updates, uncomment and set OTA password here and in credentials.cpp
 ; You can set upload_port to device's ip after it's set up for the first time
 ; Use the same password in SlimeVR Server to let it OTA Update the device
 ;upload_protocol = espota
--- a/src/GlobalVars.h
+++ b/src/GlobalVars.h
@@ -31,6 +31,8 @@
 #include "configuration/Configuration.h"
 #include "network/connection.h"
 #include "network/manager.h"
+#include "network/wifihandler.h"
+#include "network/wifiprovisioning.h"
 #include "sensors/SensorManager.h"
 #include "status/StatusManager.h"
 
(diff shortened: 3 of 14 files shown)


src/defines.h
```diff
@@ -28,7 +28,7 @@
 // Set parameters of IMU and board used
 #define IMU IMU_AUTO
 #define SECOND_IMU IMU_AUTO
-#define BOARD BOARD_SLIMEVR_V1_2
+#define BOARD BOARD_SLIMEVR
```
kounocom, line 31: Unintended?

gorbit99: Yes

Eirenliel:
Kind of a big change for the next release that will go into prod, so i think it;s better to postpone the merge.

Bluebie:
I feel uneasy about devices sharing WiFi credentials to anything that asks, over the air, without user interaction to explicitly authorise it. It's good that it's only available during active provisioning, but I think it's worth considering going further:

- Should users be asked if they want to enable this functionality so they can make their own choice about how secure they want to keep their WiFi network?
- Maybe the wired provisioning hub device should relay a list of every discovered device to the slime server UI, which could expose the device name and any other identifying info (mac addr? vendor name?), with an "Add" button or similar to explicitly authorise that radio discovered device to receive provisioning?
- Consider setting the ESP-NOW PMK and LMK security keys to something unique to slime, so casual sniffing of wifi packets with software that maybe familiar with ESP-NOW but not specifically designed to target slimes, doesn't automatically get to decrypt wifi password info.
- If firmwares can accommodate the weight, it would be lovely to include something like noise crypto or tweetnacl libraries and use asymmetric crypto to generate a unique public/private keypair and secure these provisioning messages against casual sniffing inside a unique ephemeral channel, so successful attacks would require transmitting a unique provisioning request and not just knowing the default sdk preshared keys (which i think are currently left on esp-now defaults?) or hard coded project keys (if we change the defaults in firmware)

Notably the esphome project is already using noise crypto on a wide variety of commercial IoT products very successfully across almost all of Espressif's chipsets, and also on RP2040 and a few other ultra cheap wifi chipsets, so we can be confident it will perform well on the most commonly used microcontrollers for WiFi slimes. Since WiFi Provisioning doesn't affect the smol slimes, i think we can ignore nordic semi compatibility (though in principal it should work there too, i think)

ImUrX:
I understand the authorization for sharing wi-fi password but also the communication range is also very short according to the caveats section

PR #437, opened 2025-05-05
Remove Mahony and Madgwick

gorbit99:
Mahony and madgwick (probably) aren't used by anyone and it isn't even supported by sfusion based IMU drivers. This PR therefore removes that and any mention of them from the code. It also removes SensorFusionRestDetect, merging its functionality with SensorFusion and using that instead.


Eirenliel (approved):

Eirenliel:
Plz rebase 🥺

Issue #146, opened 2022-04-04
Add ID command


Eirenliel:
An ID command sent from the server should force tracker to light up the LED for identification. Command can contain trackerId but the tracker can ignore it if there are no way to identify only extension. Command should be acknowledged by replying it back to the server. Another command with state off should stop the process and again be acknowledged to the server.

Eirenliel:
![](https://github.trello.services/images/mini-trello-icon.png) [Add ID command to tracker](https://trello.com/c/av0PCiCA/14-add-id-command-to-tracker)

TheDevMinerTV:
This should be added into the new protocol tbh, there's no need to change the current netcode just to add this and then reimplement it later.

gorbit99:
Is this still relevant?

Eirenliel:
Yes, and it will be also interesting for smol slimes

gorbit99:
@Eirenliel does it make sense for the packet to contain trackerId? Not like the extension would have a led to light up. Instead what about just a boolean for "light up led" vs "dim led"?

Eirenliel:
Eh it's better to have the tracker id and let the tracker decide what to do with it. We support trackers with many extensions and many designs.

sctanf:
if support is needed for smols, an issue can be created there

Eirenliel:
@sctanf yeah but for now we can have it in the protocol on both sides and see how we want to actually use it.

PR #438, opened 2025-05-05
Add id command

gorbit99:
Implements #146 

Adds an identification packet with id 28. This packet forces the LED on and doesn't let anything else turn it off until the server allows that. It also acknowledges this packet back to the server.

Probably a rework of the LED system to base it off priorities and whatnot would be in order to make this a bit cleaner.

The packet contains a sensorId, so it is in theory possible to idenfitify just an extension, but since this isn't currently supported in the code at all even if there were to be an LED on the extension, the param gets ignored and as a fallback the main LED gets lit up instead. Ideas are welcome.


src/network/packets.h
```diff
@@ -54,6 +54,7 @@ enum class SendPacketType : uint8_t {
 	// RotationAcceleration = 23,
 	AcknowledgeConfigChange = 24,
 	FlexData = 26,
+	AcknowledgeIdentificiation = 28,
```
Eirenliel, line 57: who ate packet 27? 😭

Eirenliel (changes requested):

Eirenliel (approved):

Eirenliel:
plz rebase 🥺

Eirenliel (approved):

Issue #275, opened 2023-08-14
V3.2 LED Blinking for IMU error overriden by WiFi Blinking?


unlogisch04:
Hm the blinking on the firmware seems to be broken. Why is it not blinking 5 times for IMU error?⁠
https://discord.com/channels/817184208525983775/878727840118505533/1140784756720685178

PR #439, opened 2025-05-06
Move SERVER_SEARCHING LED status indication to the LEDManager

SomaticTom:
Add SERVER_SEARCHING status to the StatusManager and move the server searching LED effect to the LEDManager. This allows other LED status indications, like charging and charge complete to show if the tracker is connected to WiFi but the server is not running.


src/status/Status.h
```diff
@@ -8,7 +8,8 @@ enum Status {
 	LOW_BATTERY = 1 << 1,
 	IMU_ERROR = 1 << 2,
 	WIFI_CONNECTING = 1 << 3,
-	SERVER_CONNECTING = 1 << 4
+	SERVER_SEARCHING = 1 << 4, 
+	SERVER_CONNECTING = 1 << 5
```
gorbit99, line 12: Considering how status is technically not just an internal value, but also printed to serial console sometimes, it might be a good idea to not change around the ordering

SomaticTom: Good point. I figured it was a more logical order, however. Should I change it?

gorbit99: Yeah, IMO

SomaticTom: Done

SomaticTom:
A possible fix for [#275](https://github.com/SlimeVR/SlimeVR-Tracker-ESP/issues/275)

PR #443, opened 2025-05-12
Run CI for new boards with own defines

Eirenliel:
Makes CI run for new boards, but this time it also sets defines correctly, so the FW can be used in updates, and it also gets properly tested. It also builds glove fw for fun of it, and we can just set whatever defines we want to test in the matrix and it will do it.

I also "by accident" reorganized stuff a bit...


src/defines.h
```diff
@@ -26,214 +26,36 @@
...
+#endif
+#ifndef IMU_ROTATION
 #define IMU_ROTATION DEG_270
+#endif
+#ifndef SECOND_IMU_ROTATION
 #define SECOND_IMU_ROTATION DEG_270
+#endif
 
+#ifndef PRIMARY_IMU_OPTIONAL
 #define PRIMARY_IMU_OPTIONAL false
```
gorbit99, line 46: Don't think wrapping these in ifndefs is necessary. Where else would the user define these? Honestly would just leave them there and if someone were to want to change these they can in place.

kounocom (approved):

src/board_default.h
```diff
@@ -174,3 +174,215 @@ BATTERY_R1(10)
...
+		DIRECT_WIRE(PIN_IMU_SCL, PIN_IMU_SDA), \
+		SECONDARY_IMU_OPTIONAL,                \
+		DIRECT_PIN(PIN_IMU_INT_2),             \
+		0                                      \
+	)
+#endif
+#endif
+#else
+
+// efaunlt definitions for the GLOVE
```
gorbit99, line 248:
```suggestion
// default definitions for the GLOVE
```

src/board_default.h
```diff
@@ -174,3 +174,215 @@ BATTERY_R1(10)
...
+#ifndef MAX_SENSORS_COUNT
+#define MAX_SENSORS_COUNT 2
+#endif
+#ifndef TRACKER_TYPE
+#define TRACKER_TYPE TrackerType::TRACKER_TYPE_SVR_ROTATION
+#endif
+// Set I2C address here or directly in IMU_DESC_ENTRY for each IMU used
+// If not set, default address is used based on the IMU and Sensor ID
+// #define PRIMARY_IMU_ADDRESS_ONE 0x4a
+// #define SECONDARY_IMU_ADDRESS_TWO 0x4b
```
gorbit99, line 191: Move back to defines.h maybe? Makes slightly more sense there

src/board_default.h
```diff
@@ -174,3 +174,215 @@ BATTERY_R1(10)
...
+	SENSOR_INFO_ENTRY(1, SensorPosition::POSITION_RIGHT_LITTLE_INTERMEDIATE) \
+	SENSOR_INFO_ENTRY(2, SensorPosition::POSITION_RIGHT_RING_INTERMEDIATE)   \
+	SENSOR_INFO_ENTRY(3, SensorPosition::POSITION_RIGHT_RING_DISTAL)         \
+	SENSOR_INFO_ENTRY(4, SensorPosition::POSITION_RIGHT_MIDDLE_INTERMEDIATE) \
+	SENSOR_INFO_ENTRY(5, SensorPosition::POSITION_RIGHT_MIDDLE_DISTAL)       \
+	SENSOR_INFO_ENTRY(6, SensorPosition::POSITION_RIGHT_INDEX_INTERMEDIATE)  \
+	SENSOR_INFO_ENTRY(7, SensorPosition::POSITION_RIGHT_INDEX_DISTAL)        \
+	SENSOR_INFO_ENTRY(8, SensorPosition::POSITION_RIGHT_THUMB_PROXIMAL)      \
+	SENSOR_INFO_ENTRY(9, SensorPosition::POSITION_RIGHT_THUMB_DISTAL)
+#else  // GLOVE_SDIE
```
gorbit99, line 384:
```suggestion
#else  // GLOVE_SIDE
```

src/board_default.h
```diff
@@ -174,3 +174,215 @@ BATTERY_R1(10)
...
+#error "Glove side not defined"
+#endif  // GLOVE_SDIE
```
gorbit99, line 386:
```suggestion
#endif  // GLOVE_SIDE
```

loucass003:
Maybe i am picky. But the massive if in the defines for all the gloves defines, i cannot even find the endif from github preview. Maybe we could put that whole thing into its own header?

src/defines.h
unlogisch04: they are needed for the build.py because it will redefine the board. 
I also think not the nicest solution. 

```ini
[env:BOARD_SLIMEVR_V1_2]
platform = espressif8266 @ 4.2.1
board = esp12e
build_flags =
  ${env.build_flags}
  -D BOARD=BOARD_SLIMEVR_V1_2
  ```
  
  Maybe nicer would be to have defines.h excluded if we run ci with -D NODEFINES=true ? And disable the include in globals.h? 
  So it would not affect the avarange diy person or webflasher.

PR #447, opened 2025-05-14
Allow multiple WHOAMI values per SoftFusion sensor

kounocom:
This PR adds the functionality of allowing multiple WHOAMI values per SoftFusion sensor. It is mainly for parity with the old BMI160 driver, which handled revision 3 chips properly (those have a value of 0xD3), but it also allows LSM6DS3 to work with the LSM6DS3TR-C driver.

### DIFF
--- a/src/sensors/softfusion/softfusionsensor.h
+++ b/src/sensors/softfusion/softfusionsensor.h
@@ -76,7 +76,26 @@ class SoftFusionSensor : public Sensor {
 	bool detected() const {
 		const auto value
 			= m_sensor.m_RegisterInterface.readReg(SensorType::Regs::WhoAmI::reg);
-		if (SensorType::Regs::WhoAmI::value != value) {
+		if constexpr (requires { SensorType::Regs::WhoAmI::values.size(); }) {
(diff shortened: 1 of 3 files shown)


src/sensors/softfusion/softfusionsensor.h
```diff
@@ -76,7 +76,26 @@ class SoftFusionSensor : public Sensor {
 	bool detected() const {
 		const auto value
 			= m_sensor.m_RegisterInterface.readReg(SensorType::Regs::WhoAmI::reg);
-		if (SensorType::Regs::WhoAmI::value != value) {
+		if constexpr (requires { SensorType::Regs::WhoAmI::values.size(); }) {
+			for (auto possible : SensorType::Regs::WhoAmI::values) {
+				if (value == possible) {
+					return true;
+				}
+			}
```
gorbit99, line 84: nit: could use std::find or std::any_of instead, though it's unclear if that would improve on readability

gorbit99 (approved):

ShineBrightMeow (approved):

Eirenliel (approved):

PR #452, opened 2025-05-25
Esp32 fix defines

unlogisch04:
make all esp32 define the same
only check if ESP32 is defined. The value changed from "1" to "ESP32" in pioarduino and tasmota in newer versions.
update esp32 to tasmota version


gorbit99:
This PR seems misnamed? Since it also changes the platform over to tasmota

unlogisch04:
Your right, wanted to make a second branch for it.

gorbit99 (approved):

PR #454, opened 2025-05-26
Fix sensor count reporting

gorbit99:
Currently main, when not using IMU_AUTO, will report the configured sensor count ("n sensor(s) configured") as the number of entries in the SENSOR_DESC_LIST, even if some of them fail to be configured. This PR fixes that.


Eirenliel:
I'm not sure this is a correct behavior. If there are ERRORS it still should be like sensor is attached. It should only be not a sensor if the sensor isn't found, which isn't what isWorking() checks.

gorbit99:
@Eirenliel this PR just restores the behaviour of the code from before code cleanup was merged.
https://github.com/SlimeVR/SlimeVR-Tracker-ESP/blob/c6e3e6247f6a4264380740cd25a3ecbfc78feff7/src/sensors/SensorBuilder.cpp#L71
I think it is the correct behaviour, as a sensor shouldn't really be considered "configured" if it couldn't be communicated with.

PR #461, opened 2025-06-12
Send vendor information on handshake

Eirenliel:
Embed vendor information into build parameters & defines so SlimeVR vendors can maintain their own firmware updates without adding new boards

Bump protocol version to 21

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -68,6 +68,14 @@ board = esp12e
 ; Comment out this line below if you have any trouble uploading the firmware
 ; and if it has a CP2102 on it (a square chip next to the usb port): change to 3000000 (3 million) for even faster upload speed
 upload_speed = 921600
+build_flags =
+  ${env.build_flags}
+  -D BOARD=BOARD_SLIMEVR_V1_2
+  -D VENDOR_NAME='"SlimeVR"'
+  -D VENDOR_URL='"https://slimevr.dev"'
+  -D PRODUCT_NAME='"SlimeVR Tracker v1.2"'
+  -D UPDATE_ADDRESS='"SlimeVR/SlimeVR-Tracker-ESP"'
+  -D UPDATE_NAME='"BOARD_SLIMEVR_V1_2-firmware"'
 
 ; Uncomment below if you want to build for ESP-01
 ;[env:esp01_1m]
--- a/platformio.ini
+++ b/platformio.ini
@@ -68,14 +68,6 @@ board = esp12e
 ; Comment out this line below if you have any trouble uploading the firmware
 ; and if it has a CP2102 on it (a square chip next to the usb port): change to 3000000 (3 million) for even faster upload speed
 upload_speed = 921600
-build_flags =
-  ${env.build_flags}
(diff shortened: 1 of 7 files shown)


Eirenliel:
This is for future-proofing, so the server side isn't implemented yet.

TheDevMinerTV (approved):
lgtm

0forks (approved):

ShineBrightMeow (approved):

platformio.ini
```diff
@@ -68,6 +68,14 @@ board = esp12e
...
 ; and if it has a CP2102 on it (a square chip next to the usb port): change to 3000000 (3 million) for even faster upload speed
 upload_speed = 921600
+build_flags =
+  ${env.build_flags}
+  -D BOARD=BOARD_SLIMEVR_V1_2
+  -D VENDOR_NAME='"SlimeVR"'
+  -D VENDOR_URL='"https://slimevr.dev"'
+  -D PRODUCT_NAME='"SlimeVR Tracker v1.2"'
+  -D UPDATE_ADDRESS='"SlimeVR/SlimeVR-Tracker-ESP"'
+  -D UPDATE_NAME='"BOARD_SLIMEVR_V1_2-firmware"'
```
unlogisch04, line 78: If someone is building from platformio.ini it is most likely a diy tracker.
At least for final we should change the values back to something that allows us to differentiate DIY and the normal Officials

```suggestion
build_flags =
  ${env.build_flags}
  -D VENDOR_NAME='"DIY"'
  -D VENDOR_URL='"https://docs.slimevr.dev"'
  -D PRODUCT_NAME='"DIY SlimeVR trackers"'
  -D UPDATE_ADDRESS='""'
  -D UPDATE_NAME='""'
```


Also the the following fields should be a described, how they have to be formated:
VENDOR_NAME, VENDOR_URL, PRODUCT_NAME, **UPDATE_ADDRESS, UPDATE_NAME**

But also do we need to set it here? Defaults would work too

src/globals.h
```diff
@@ -52,3 +52,23 @@
...
+#ifndef VENDOR_NAME
+#define VENDOR_NAME "Unknown"
+#endif
+
+#ifndef VENDOR_URL
+#define VENDOR_URL ""
+#endif
+
+#ifndef PRODUCT_NAME
+#define PRODUCT_NAME "SlimeVR Tracker"
```
unlogisch04, line 65:
```suggestion
#define PRODUCT_NAME "DIY SlimeVR Tracker"
```

unlogisch04 (changes requested):
In general it seems good.
I think the following should be changed:
- platformio.ini -> The default should be not the Branding for the Official trackers. Else everyone using VSCode to compile his custom boards has SlimeVR Tracker.
- serial output-> would be great todo something with this values. (for support)
  - main.cpp Starting up message?
  - serialcommands.cpp printState()?
  - serialcommands.cpp test ?

platformio.ini
Eirenliel: I was not planning to keep it in, this piece got here by accident

PR #464, opened 2025-07-10
Add IMU_USE_EXTERNAL_CLOCK to debug.h

kounocom:
Allows user to easily disable external IMU clock, if they are using a design which doesn't have one


gorbit99 (approved):
lgtm

ButterscotchV (approved):

PR #473, opened 2025-09-30
WiFi Fix

gorbit99:
This PR fixes 2 issues with #435 

1. The constant printing of "." was present even once the connection process failed completely, this was reverted to the original behaviour
2. Apparently ESP32 can't fetch the WiFi SSID without WiFi.begin() being called first, I applied a fix

### DIFF
--- a/src/network/wifihandler.cpp
+++ b/src/network/wifihandler.cpp
@@ -21,7 +21,10 @@
 	THE SOFTWARE.
 */
 #include "GlobalVars.h"
+#include "WiFi.h"
+#include "esp_wifi_types.h"
 #include "globals.h"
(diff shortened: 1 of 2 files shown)


src/network/wifihandler.cpp
```diff
@@ -62,14 +65,13 @@ IPAddress WiFiNetwork::getAddress() { return WiFi.localIP(); }
 
 void WiFiNetwork::setUp() {
 	// Don't need to save the already saved credentials or the hardcoded ones
```
ButterscotchV, line 67: Probably remove this comment too

ButterscotchV (approved):

TheDevMinerTV (approved):

PR #478, opened 2025-10-23
Implement mDNS discovery

gorbit99:
This PR implements mDNS discovery. It searches for an mDNS record with host name "slimevr-server.local" and resolves it into the server's ip address. In some specific cases this should function better than broadcast based discovery.

mDNS packets are implemented manually to both make the process non-blocking and have it peacefully coexist with broadcast based discovery.

### DIFF
--- a/src/network/connection.cpp
+++ b/src/network/connection.cpp
@@ -21,6 +21,8 @@
 	THE SOFTWARE.
 */
 
+#include "connection.h"
+
 #include <cstdint>
 #include <cstring>
 #include <ratio>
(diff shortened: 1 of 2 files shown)


gorbit99:
@Eirenliel this currently queries A records, which are unique across a network, when the second server would attempt to register itself with the same hostname, it would fail and potentially rename itself. This can be used to detect instances of the server already running on the network, which could be used to implement https://github.com/SlimeVR/SlimeVR-Server/issues/1581

In theory this could be solved by querying for SRV records instead, which can return all of them, but I didn't see the point, since we'd just use the first one that responds either way.

PR #491, opened 2025-11-17
fix no broadcast after disconnect

unlogisch04:
After a disconnect the trackers did not send a broadcast to discover the server but did directly try to send it to the server. So if for some reason the server did change its IP address the tracker had to be rebooted.


platformio.ini
loucass003: unrelated changes?

loucass003: i would make it a separate pr. All tho its prob fine even rn who cares

Eirenliel (approved):

PR #493, opened 2025-11-18
fix icm42688

unlogisch04:
Try to fix the ICM42688
Currently with that configuration it still snaps around when printing stuff or saving calibration.

### DIFF
--- a/src/sensors/softfusion/drivers/icm42688.h
+++ b/src/sensors/softfusion/drivers/icm42688.h
@@ -56,10 +56,12 @@ struct ICM42688 {
 	// corresponding sensitivity scale factor values are 131 LSB/dps for gyroscope and
 	// 8192 LSB/g for accelerometer.
 	static constexpr float GyroSensitivity = (DEBUG_ICM42688_HIRES ? 131.0f : 32.8f);
-	static constexpr float AccelSensitivity = (DEBUG_ICM42688_HIRES ? 8192.0f : 4096.0f);
+	static constexpr float AccelSensitivity
+		= (DEBUG_ICM42688_HIRES ? 8192.0f : 4096.0f);
 
 	static constexpr float TemperatureBias = 25.0f;
-	static constexpr float TemperatureSensitivity = (DEBUG_ICM42688_HIRES ? 132.48f : 2.07f);
+	static constexpr float TemperatureSensitivity
+		= (DEBUG_ICM42688_HIRES ? 132.48f : 2.07f);
 	static constexpr float TemperatureZROChange = 20.0f;
 
 	static constexpr VQFParams SensorVQFParams{};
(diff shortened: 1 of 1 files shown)


src/sensors/softfusion/drivers/icm42688.h
```diff
@@ -113,7 +134,7 @@ struct ICM42688 {
 		static constexpr uint8_t FifoCount = 0x2e;
 		static constexpr uint8_t FifoData = 0x30;
 	};
-
+#if debug_icm42688_20bit
```
gorbit99: Please define both structs as normal, but named FifoEntryAlignedDefault and FifoEntryAlignedHires or something similar, then you can create an alias with
```cpp
using FifoEntryAligned = std::conditional<DEBUG_ICM42688_HIRES, FifoEntryAlignedHires, FifoEntryAlignedDefault>;

File: src/sensors/softfusion/drivers/icm42688.h [hunk 2]
```diff
@@ -170,23 +211,47 @@ struct ICM42688 {
...
-					static_cast<int32_t>(entry.part.accel[1]) << 4
-						| (static_cast<int32_t>(entry.part.ylsb) & 0xf0 >> 4),
-					static_cast<int32_t>(entry.part.accel[2]) << 4
-						| (static_cast<int32_t>(entry.part.zlsb) & 0xf0 >> 4),
+					static_cast<int32_t>(entry.part.accel[0]) << 2
+						| (static_cast<int32_t>(entry.part.xlsb) & 0xf0 >> 6),
+					static_cast<int32_t>(entry.part.accel[1]) << 2
+						| (static_cast<int32_t>(entry.part.ylsb) & 0xf0 >> 6),
+					static_cast<int32_t>(entry.part.accel[2]) << 2
+						| (static_cast<int32_t>(entry.part.zlsb) & 0xf0 >> 6),
```
gorbit99: `>>` has higher precedence than `&`, this left right shifts the 0xf0 constant first then applies that as a mask.

src/sensors/softfusion/drivers/icm42688.h
```diff
@@ -170,23 +211,47 @@ struct ICM42688 {
...
+			const int32_t gyroData[3]{
+				static_cast<int32_t>(entry.part.gyro[0]),
+				static_cast<int32_t>(entry.part.gyro[1]),
+				static_cast<int32_t>(entry.part.gyro[2]),
+			};
+#endif
 			callbacks.processGyroSample(gyroData, GyrTs);
 
 			if (entry.part.accel[0] != -32768) {
+#if debug_icm42688_20bit
```
gorbit99: Same as above

Eirenliel (approved):

PR #499, opened 2025-11-27
Fix LSM* Accelerometer Sensitivity

kounocom:
Fixes accelerometer sensitivity for LSM* IMUs since I broke it when changing scale to 4G


unlogisch04:
Checked on LSMDSO accel is fixed

unlogisch04 (approved):
lgtm

PR #507, opened 2025-12-19
(rework) support for TTGO T-Display board

wafflekirima:
update all declare define of TTGO T-Display board and testing with ICM-45686(ICM-45686 with QMC6309 breakout board) are working fine.

also define consts board id and board default pinout

image
image


board-defaults.json
```diff
@@ -585,6 +585,45 @@
...
+					"shieldR": 0,
+					"pin": "34"
+				},
+				"LED": {
+					"LED_PIN": "2",
+					"LED_INVERTED": true
+				}
+			},
+			"flashingRules": {
+				"applicationOffset": 0,
```
loucass003: Please make sure the offset is correct. 
it is the address where the firmware should be flashed in the eeprom. 
if i recall the value on esp32 is not 0

PR #511, opened 2026-01-11
Feat flashmode for ESP8266

unlogisch04:
On Official Slime HW it is not possible to flash them just over USB when they are in a boot loop. 
They are based from ESP8266. 
~~Left ESP32* out of loop as they either have a HW solution to put them into Flash Mode or use USB JTag on the chip.~~
Target is, that there is a fail safe at begin of the program so it goes into flash mode if it reboots more than 3 times after a crash. For ESP32 it is directly after a crash. For HW with softpower solution (gorbits as example) the code in init has maybe to be changed, so a reboot with a button press is still possible.

Current Status is:
- A safe mode shell
- Boot Prompt with Infos from Flash
- Small Serial command set:
  - FRST - Factory Reset, cleans the config file and WiFi credentials (useful by crashes to wrong configs)
  - SET FLASHMODE - Sets the tracker into flashmode does work on ESP8266 and ESP32-C3 )
  - REBOOT - Reboots the tracker
- Still not able to recover from updates that have been interrupted by power off at the wrong time

Tasks:
- [x] Needs to be checked if this constructor runs rly before all other constructors.
- [x] Testing with official Slimes
- [x] Testing that no breaks for ESP32 variants. 
- [x] Lint

### DIFF
--- a/src/configuration/Configuration.cpp
+++ b/src/configuration/Configuration.cpp
@@ -38,10 +38,30 @@
 #define DIR_TOGGLES_OLD "/toggles"
 #define DIR_TOGGLES "/sensortoggles"
 
+// Global variable defined at file scope (not in namespace) so init.h can access it
+extern bool g_safeModeDeferredFactoryResetRequested;
+
 namespace SlimeVR::Configuration {
 void Configuration::setup() {
 	if (m_Loaded) {
 		return;
+	}
+
+	// Safe mode defers LittleFS formatting to normal runtime for reliability.
+	// If requested, perform the filesystem format before config load.
+	if (g_safeModeDeferredFactoryResetRequested) {
+		m_Logger.warn("Deferred safe-mode factory reset detected, formatting LittleFS");
+		g_safeModeDeferredFactoryResetRequested = false;
+
+		LittleFS.begin();
+		if (!LittleFS.format()) {
+			m_Logger.error("Deferred LittleFS format failed");
+		} else {
+			m_Logger.info("Deferred LittleFS format OK");
+			m_Logger.warn("Rebooting after deferred safe-mode factory reset");
+			ESP.restart();
+			return;
+		}
 	}
 
 	bool status = LittleFS.begin();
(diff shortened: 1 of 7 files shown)


unlogisch04:
Target is, that there is a fail safe at begin of the program so it goes into flash mode if it reboots more than 3 times (for esp8266) and for ESP32 at the first reboot.

Currently it goes directly into flashmode, or stops the execution to wait for jtag. (Or to give the user chance to upload the firmware on esp32c3) 
Not sure if it would be better to present a cmd line, where you could 
1. factory reset 
2. set the tracker to flash mode
3. delete some config files

unlogisch04:
Where i did use ai in the pr:
Generaly for comments (autocomplete)
Print formating.
Structure Build up.

(FactoryResetRequest) <https://github.com/SlimeVR/SlimeVR-Tracker-ESP/pull/511/changes#diff-8f48258f1ee9093c323d5ec2f0c5d67b1e9e92b271e6f2a3080fc3929f7ecbaeR50-R66>
<https://github.com/SlimeVR/SlimeVR-Tracker-ESP/pull/511/changes#diff-c107d11265fc4260eff0cb82da5ca8373e7b12d6d8f73b0e59245badc5a884b2R33>

just was to lazy to write all by hand: <https://github.com/SlimeVR/SlimeVR-Tracker-ESP/pull/511/changes#diff-9f367125182c6d0af13cc3ee58f08ddd36f16b6be4e1abda8b1921ba2c70a223R104-R166>

Generally the Output formatting of the println. (like <https://github.com/SlimeVR/SlimeVR-Tracker-ESP/pull/511/changes#diff-6fb020411638e373ee8f515c197b3cea94fd1191bb5a0204ea6db1df2692a49eR145-R154>)

And the loop here: (loop structure, handling of serial input)
<https://github.com/SlimeVR/SlimeVR-Tracker-ESP/pull/511/changes#diff-6fb020411638e373ee8f515c197b3cea94fd1191bb5a0204ea6db1df2692a49eR172-R210>

PR #515, opened 2026-02-20
Fix the order of vqf updates to the recommended order.

ksy2645:
The text may be awkward due to the use of a translator.

Issue: If the order is wrong, this will cause an issue where it updateAcc with the gyrQuat of the previous sample.

Reference:
  - [[VQF DOC - VQF class]](https://vqf.readthedocs.io/en/stable/ref_cpp.html#)

~~# Test Setup~~
~~**Device:** Chest tracker (LSM6DSR + MMC5603NJ)~~
~~**Magnetometer: OFF**~~

~~Gyro bias calibration logs:~~
```
[INFO ] [LSM6DSR:0] Calibrated gyro bias at 22.921875C: 10.919210 -13.017056 2.500000
[INFO ] [LSM6DSR:0] Calibrated gyro bias at 36.132812C: 11.463371 -14.227714 5.920565
```

~~# Test Method~~
~~1. Prepare a chest tracker with 2-point temperature gyro bias calibration (22.9°C and 36.13°C as shown above).~~
~~2. Power on the tracker and leave it resting flat for 30 seconds~~
~~  - Purpose: allow “rest” handling and any remaining gyro bias calibration to settle.~~
~~3. Wear it directly on bare skin and stay active for 30 minutes~~
~~  - Purpose: stabilize strap tension and bring the tracker to a realistic operating temperature.~~
~~4. Perform Full Reset → Body Reset.~~
~~5. Record and observe drift for 30 minutes on the Body Proportions screen.~~

~~# Result~~
~~Compared to the previous behavior, drift was reduced by approximately 1–2 degrees.~~

~~This isn’t a perfectly controlled test (human-body bias, strap tension, environment, etc.), but even considering those factors, the improvement looked significant in my use case.~~

~~I’m sharing this because I think it could also be beneficial to apply the same fix to the SlimeVR Tracker firmware.~~

BEFORE FIX
AFTER FIX

### DIFF
--- a/src/sensors/SensorFusion.cpp
+++ b/src/sensors/SensorFusion.cpp
@@ -7,8 +7,8 @@ void SensorFusion::update6D(
 	sensor_real_t Gxyz[3],
 	sensor_real_t deltat
 ) {
-	updateAcc(Axyz, deltat);
 	updateGyro(Gxyz, deltat);
+	updateAcc(Axyz, deltat);
 }
 
 void SensorFusion::update9D(
@@ -17,9 +17,9 @@ void SensorFusion::update9D(
 	sensor_real_t Mxyz[3],
 	sensor_real_t deltat
 ) {
-	updateMag(Mxyz, deltat);
-	updateAcc(Axyz, deltat);
 	updateGyro(Gxyz, deltat);
+	updateAcc(Axyz, deltat);
+	updateMag(Mxyz, deltat);
 }
 
 void SensorFusion::updateAcc(const sensor_real_t Axyz[3], sensor_real_t deltat) {


unlogisch04:
From my point of view: Change request is valid but it does not affect the sensor you did mention.

While i agree that this is technically wrong: comments on the source file: https://github.com/SlimeVR/SlimeVR-Tracker-ESP/blob/main/lib/vqf/vqf.h#L718-L748

I don't see any connection from this function to LSM6DSR.
It gets read from: https://github.com/SlimeVR/SlimeVR-Tracker-ESP/blob/main/src/sensors/softfusion/drivers/lsm6dsr.h#L118-L125

https://github.com/SlimeVR/SlimeVR-Tracker-ESP/blob/main/src/sensors/softfusion/drivers/lsm6ds-common.h#L57-L108
Where the order depends on the fifo:
```cpp
		std::array<uint8_t, FullFifoEntrySize * 8> read_buffer;  // max 8 readings
		const auto bytes_to_read = std::min(
									   static_cast<size_t>(read_buffer.size()),
									   static_cast<size_t>(fifo_bytes)
								   )
								 / FullFifoEntrySize * FullFifoEntrySize;
		m_RegisterInterface
			.readBytes(Regs::FifoData, bytes_to_read, read_buffer.data());
		for (auto i = 0u; i < bytes_to_read; i += FullFifoEntrySize) {
			FifoEntryAligned entry;
			uint8_t tag = read_buffer[i] >> 3;
			memcpy(
				entry.raw,
				&read_buffer[i + 0x1],
				sizeof(FifoEntryAligned)
			);  // skip fifo header

			switch (tag) {
				case 0x01:  // Gyro NC
					callbacks.processGyroSample(entry.xyz, GyrTs);
					break;
				case 0x02:  // Accel NC
					callbacks.processAccelSample(entry.xyz, AccTs);
					break;
				case 0x03:  // Temperature
					callbacks.processTempSample(entry.xyz[0], TempTs);
					break;
			}
		}
```

the callback leads then to https://github.com/SlimeVR/SlimeVR-Tracker-ESP/blob/main/src/sensors/softfusion/softfusionsensor.h#L122-L131
```cpp
	void processGyroSample(const RawSensorT xyz[3], const sensor_real_t timeDelta) {
		sensor_real_t gyroData[]
			= {static_cast<sensor_real_t>(xyz[0]),
			   static_cast<sensor_real_t>(xyz[1]),
			   static_cast<sensor_real_t>(xyz[2])};
		calibrator.scaleGyroSample(gyroData);
		m_fusion.updateGyro(gyroData, calibrator.getGyroTimestep());

		calibrator.provideGyroSample(xyz);
	}
```
The line calling the vqf in the updateGyro from Sensorfusion.cpp is the `m_fusion.updateGyro(gyroData, calibrator.getGyroTimestep());`

ksy2645:
I've verified that the code doesn't use that function. This is my mistake. Don't worry about the test results. I'll close this pull request.

PR #518, opened 2026-03-04
ICM-55686 Implementation

gorbit99:
This PR adds a firmware implementation for the ICM-55686 sensor. It heavily mirrors the ICM-45 implementation with some small differences, like the register addresses (and sometimes values) and the endianness of the FIFO being different.


kounocom:
@gorbit99 enable AAF and interpolator and I'll approve

Eirenliel (approved):

PR #523, opened 2026-03-25
Increase MAX_PACKET_SIZE from 128 to 300 for BNO085 SPI

halfsohappy:
## Summary

- BNO085 can return packets up to 270 bytes during sensor hub initialisation
- `MAX_PACKET_SIZE 128` sized `shtpData[]` at only 128 bytes, so the SPI receive loop wrote up to 270 bytes into it
- This corrupted the adjacent heap and triggered `multi_heap_poisoning` asserts on ESP32-S3 at boot (`CORRUPT HEAP` panic at `0x3fcb1dbc`)
- Increase to 300 (safely above the 270-byte maximum) so the buffer is never overflowed

## Test plan

- [ ] Flash ab7 board (ESP32-S3, BNO085 on SPI) and confirm no `CORRUPT HEAP` panic at boot
- [ ] Verify BNO085 quaternion data flows normally after init

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Eirenliel:
💀

PR #526, opened 2026-04-16
Fix typo in bno080sensor.cpp logging

ishotjr:


ScaleShiftX (approved):
Incredible work. Well-spotted. This exactly the kind of integrity we need in this repo. Your dedication to discovering, uncovering, disseminating, and inseminating even the smallest issues, even the non-user-facing issues, is paramount to the continuing excellence of this repository.

Before merging this, I would request that go the extra kilometre here while we're at it, and improve our use of language further than just correcting typos - we have an opportunity here to reduce duplicated phrasing. In this log, the word 'status' is used twice. I humbly suggest the following:

"BBNO$ calibration received: Status: %d, Accel: %d, Gyro: %d, "

Thus the duplicated phrasing bathwater is eliminated along with the typo baby.

Once again, well done, and kudos to you.

PR #530, opened 2026-05-31
Refactor build.py to improve platformio command resolution and path h…

unlogisch04:
…andling

This commit fixes that pio or platformio was not found while running the build.py script on my windows This changes where done by copilot.  
Tested by me on Windows and on gh ci not on web build, but this file is not used there so should be ok.

### DIFF
--- a/ci/build.py
+++ b/ci/build.py
@@ -2,7 +2,10 @@ import json
 import os
 import sys
 import shutil
+import subprocess
+import importlib.util
 from enum import Enum
+from pathlib import Path
 from textwrap import dedent
 from typing import List
 import configparser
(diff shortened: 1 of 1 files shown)


Eirenliel:
> This changes where done by copilot.

i immediately don't trust this code.

unlogisch04:
Well agree. But im not even sure why the old code did begin to not work on my os. 
I have maybe a messed up env path. But also after reinstalling the IDE + Python was not working.
So maybe it is not needed.

Eirenliel:
Then this is enough to close the issue. Per our old policy AI-generated code is not allowed, per our new policy (pending discussion), you're not allowed to submit code that you don't know how it works.

PR #531, opened 2026-06-01
Wifiscan add channel to the output, format output

unlogisch04:
This might break some pr that try to read the SSID from the tracker over Serial as it does no longer use tab to hold the distance


Eirenliel (approved):

Eirenliel:
Is this parsed by server @loucass003 ? Just in case it breaks anything...

PR #532, opened 2026-06-06
Performance measurement

unlogisch04:
Add better performance measurement that allows to monitor where process times gets used. 
It is a rewrite by @gorbit99  of https://github.com/SlimeVR/SlimeVR-Tracker-ESP/tree/feat_benchmark from me.

It should help to watch for changes better or worse when testing new code in performance/time used in the loop.
By default it is disabled, as the printout uses some time too.


src/sensors/SensorManager.cpp
```diff
@@ -23,10 +23,19 @@
...
+#include <array>
+
+#include "../debugging/Benchmark.h"
 #include "SensorBuilder.h"
 
 namespace SlimeVR::Sensors {
 
+std::array<SlimeVR::Debugging::Benchmark, 2> sensorLoopBMs{
+	SlimeVR::Debugging::Benchmark{"IMU1 Sensor loop"},
+	SlimeVR::Debugging::Benchmark{"IMU2 Sensor loop"}
```
Eirenliel, line 35: There could be more than 2 sensor on one MCU, technically up to 255, but we do have a glove option that uses 10 (even though the results are mid, this is hardware limitation, not firmware)

Eirenliel (changes requested):

src/sensors/SensorManager.cpp
unlogisch04: Ok it could make a out of index error there. 
Is it ok just to fix the out of bound read?
To make it dynamic for MAX_SENSORS_COUNT is at least for me not possible.
@gorbit99 you know more of templating stuff, so maybe you have a solution.

Eirenliel: It's probably fine if it just doesn't crash/

Eirenliel (approved):

PR #533, opened 2026-06-07
Add Serial buffer for performance improvement

unlogisch04:
this PR goes on top of #532 
Serial Writes did take up to 54ms to write out the SerialCommands::printState();

After digging deeper i found out that the esp8266 frameworks blocks till all chars are printed out.
In taking advantage of the HW Buffer (128Bytes), to only write the set amount of chars to the buffer the blocking behavior can be avoided. (from 54 ms to about ~400us each cycle)
Arduino has the free bytes to write in the function called `Serial.availableForWrite()`

With this change we can avoid long blocking serial operation. 

the Logger function to print Arrays has been removed, and the only instance used converted.

There are many places Serial.Print.. is still used, that might show out of place:
- ota,cpp "[NOTICE] OTA updates allowed", "[NOTICE] OTA updates disabled by timeout, this is not an error"
- wifihandler.cpp (the ".")
- serialcommands.cpp (only 1 instance for get config)
- i2cscan (all the places)


Eirenliel (approved):

Eirenliel:
Please rebase

gorbit99:
~~Before this is merged, I want to make sure the buffer is flushed even in the setup method of the code, at the moment it's not doing that~~

unlogisch04:
good catch 👍

Eirenliel (approved):

PR #538, opened 2026-07-28
fix i2c scann stalling on ESP32* boards

unlogisch04:
- Move Wire.end() to the correct position
- ESP32-C6 Pin 0 and 9 Report all addresses
- Fix wrong numbers/names in output

### DIFF
--- a/lib/i2cscan/i2cscan.cpp
+++ b/lib/i2cscan/i2cscan.cpp
@@ -18,10 +18,12 @@ namespace I2CSCAN {
 		ScanState scanState = ScanState::IDLE;
     	uint8_t currentSDA = 0;
     	uint8_t currentSCL = 0;
+		uint8_t currentSDAPortIndex = 0;
+		uint8_t currentSCLPortIndex = 0;
     	uint8_t currentAddress = 1;
     	bool found = false;
 		uint8_t txFails = 0;
-    	std::vector<uint8_t> validPorts;
+    	std::vector<uint8_t> validPortsIndex;
 
 #ifdef ESP8266
 		std::array<uint8_t, 7> portArray = {16, 5, 4, 2, 14, 12, 13};
@@ -34,7 +36,7 @@ namespace I2CSCAN {
 #elif defined(ESP32C6)
 		std::array<uint8_t, 20> portArray = {0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 14, 15, 18, 19, 20, 21, 22, 23};
 		std::array<std::string, 20> portMap = {"0", "1", "2", "3", "4", "5", "6", "7", "8", "9", "10", "11", "14", "15", "18", "19", "20", "21", "22", "23"};
-		std::array<uint8_t, 5> portExclude = {12, 13, 16, 17, LED_PIN};
+		std::array<uint8_t, 7> portExclude = {0, 9, 12, 13, 16, 17, LED_PIN};
 #elif defined(ESP32)
 		std::array<uint8_t, 16> portArray = {4, 13, 14, 15, 16, 17, 18, 19, 21, 22, 23, 25, 26, 27, 32, 33};
 		std::array<std::string, 16> portMap = {"4", "13", "14", "15", "16", "17", "18", "19", "21", "22", "23", "25", "26", "27", "32", "33"};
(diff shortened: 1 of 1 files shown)


unlogisch04:
Still has some flaws in it.

unlogisch04:
Done changing stuff. 
- Now the correct ports are shown  
- On ESP32 changing i2c pins fixed (multiple positive hits till the SDA was changed)
- Added what number represents SDA / SCL

I decided to leave my comments and "debug" prints in.

Tested with:
- ESP8266 D1 Mini
- SlimeVR 1.0 R11 Board
- ESP32-C3 Beetle
- ESP32-C6 Beetle
- ESP32

- IMU MPU9250, BNO085, ICM42668

@gorbit99 can you take a look at it


For later, we should move that code, so it uses the <https://github.com/SlimeVR/SlimeVR-Tracker-ESP/blob/main/src/sensorinterface/I2CPCAInterface.h> if possible.

lib/i2cscan/i2cscan.cpp
```diff
@@ -34,41 +47,69 @@ namespace I2CSCAN {
...
-				return true;
-			}
+			activeSDAPin = portArray[sdaPortIndex];
+			activeSCLPin = portArray[sclPortIndex];
+//			Serial.printf_P(PSTR("[INFO ] [I2CSCAN] Change I2C to SDA: %d, SCL: %d\r\n"), (int)portArray[sdaPortIndex], (int)portArray[sclPortIndex]);
+		}
 
-			currentSCL = 0;
-			currentSDA++;
+		void incvalidPortsIndex(uint8_t &index) {
```
gorbit99: I'd personally write out "inc" as "increment"ú, admittedly I read "invalid" here and got very confused by the implementation

lib/i2cscan/i2cscan.cpp
```diff
@@ -34,41 +47,69 @@ namespace I2CSCAN {
...
+			index = (index+1) % validPortsIndex.size();
+		}
 
-			if (currentSDA >= validPorts.size()) {
+		bool incSDA(){
```
gorbit99: Same here as above

lib/i2cscan/i2cscan.cpp
```diff
@@ -34,41 +47,69 @@ namespace I2CSCAN {
...
+					if (!incSDA()) {
+						return false;
+					}
+				}
+				if (currentSCLPortIndex != currentSDAPortIndex) {
+					currentSCL = validPortsIndex[currentSCLPortIndex];
+					currentSDA = validPortsIndex[currentSDAPortIndex];
+// Debug
+//					Serial.printf("CCCC currentSDAPortIndex: %d currentSCLPortIndex: %d validPortsIndex.size: %d\r\n",
+//						currentSDAPortIndex, currentSCLPortIndex, validPortsIndex.size());
```
gorbit99, line 106: Remove if not needed

lib/i2cscan/i2cscan.cpp
```diff
@@ -90,53 +131,80 @@ namespace I2CSCAN {
...
+			// This will lead to problematic behavior if the Arrays are not correctly defined.
+			// As it will not able to reset back to the previous configured pins
+			return;
+		}
+
+// Debug
+//		for (const auto& portsIndex : validPortsIndex) {
+//			Serial.printf("Pin Index: %2d PinNum: %2d PinName: %s\r\n", portsIndex, portArray[portsIndex], portMap[portsIndex].c_str());
+//		}
+//		Serial.printf("startSDAPortIndex: %2d startSCLPortIndex: %2d\r\n", startSDAPortIndex, startSCLPortIndex);
```
gorbit99, line 175: Remove if not needed

lib/i2cscan/i2cscan.cpp
```diff
@@ -90,53 +131,80 @@ namespace I2CSCAN {
...
+		for (uint8_t i=0; i < validPortsIndex.size(); i++) {
+			if (portArray[validPortsIndex[i]] == defaultSDAPin) {
+				startSDAPortIndex = i;
+			}
+			if (portArray[validPortsIndex[i]] == defaultSCLPin) {
+				startSCLPortIndex = i;
 			}
 		}
 
+//		Serial.printf_P(PSTR("[ERROR] [I2CSCAN] Default I2C Ports SDA: %d SCL: %d\r\n"), defaultSDAPin, defaultSCLPin);
```
gorbit99, line 161: Remove if not needed

lib/i2cscan/i2cscan.cpp
```diff
@@ -90,53 +131,80 @@ namespace I2CSCAN {
...
+		currentSDA = validPortsIndex[currentSDAPortIndex];
+		currentSCL = validPortsIndex[currentSCLPortIndex];
+		activeSDAPin = portArray[currentSDA];
+		activeSCLPin = portArray[currentSCL];
         currentAddress = 1;
 		txFails = 0;
         scanState = ScanState::SCANNING;
+// Debug
+//		Serial.printf("AAAA currentSDAPortIndex: %d currentSCLPortIndex: %d validPortsIndex.size: %d\r\n",
+//			currentSDAPortIndex, currentSCLPortIndex, validPortsIndex.size());
```
gorbit99, line 190: Remove if not needed

gorbit99 (changes requested):

lib/i2cscan/i2cscan.cpp
unlogisch04: I would love to leave it in. For normal use it is not needed. But if you need to debug the code it is already here.

lib/i2cscan/i2cscan.cpp
unlogisch04: I would love to leave it in. For normal use it is not needed. But if you need to debug the code it is already here.

lib/i2cscan/i2cscan.cpp
unlogisch04: I would love to leave it in. For normal use it is not needed. But if you need to debug the code it is already here.

lib/i2cscan/i2cscan.cpp
unlogisch04: I would love to leave it in. For normal use it is not needed. But if you need to debug the code it is already here.

lib/i2cscan/i2cscan.cpp
unlogisch04: Ah ok will change.

unlogisch04:
@gorbit99 can you recheck?

Eirenliel (approved):

Issue #537, opened 2026-07-24
SPI-only ESP32-C3 halts on I2C initialization during boot


Feather-P:
Running an LSM6DSV over SPI on an ESP32-C3. No I2C sensors are defined in the build.

## Issue

The board hangs during boot. After tracing the startup logs, I found that it gets stuck in the global I2C reset/initialization path in `setup()`:

* `I2CSCAN::clearBus()`
* `Wire.begin()`

From the nearby comment, this code appears to be a workaround for the ESP32-C3 sometimes having its only I2C interface already open for an unknown reason:

```cpp
// For some unknown reason the I2C seem to be open on ESP32-C3 by default. Let's
// just close it before opening it again. (The ESP32-C3 only has 1 I2C.)
```

Since this board is not wired for I2C, the initialization path fails or hangs. This seems inconsistent with the bus-aware configuration already supported later by `DIRECT_SPI` and `SensorBuilder`.

There is also a FIXME in the codebase that appears to describe the similar problem:

> FIXME fix the I2C Scanner so it use the sensor list and not be called when no I2C sensor

## Local patch

For now, I am guarding the boot-time I2C calls by checking whether valid I2C pins are configured:

```cpp
if (PIN_IMU_SDA != 255 && PIN_IMU_SCL != 255) {
    // I2C clear/init logic
}
```

I have had to add similar checks around several I2C-related calls, not just `I2CSCAN::clearBus()`.

This gets the board through startup, but repeating the same pin check in multiple places feels more like a workaround than a proper fix.

## Fix, maybe

I think the startup logic should derive its behavior from the generated sensor configuration rather than assuming that I2C is always present.

### 1. Gate the I2C startup path

The preprocessor could generate a simple flag such as `HAS_I2C_SENSORS` from the configured sensor list.

The early startup path could then be gated explicitly:

```cpp
#if HAS_I2C_SENSORS
    I2CSCAN::clearBus(PIN_IMU_SDA, PIN_IMU_SCL);
    Wire.begin(PIN_IMU_SDA, PIN_IMU_SCL);
#endif
```

The exact implementation does not necessarily need to use a preprocessor flag, but the decision should ideally come from the same sensor topology represented by `SENSOR_DESC_LIST`.

### 2. Make the shared SPI configurable

On the ESP32-C3, routing SPI to non-default pins requires calling `SPIClass::begin()` with explicit pins:

```cpp
SPI.begin(sck, miso, mosi, -1);
[...]

PR #539, opened 2026-07-29
Add configurable shared SPI bus pins

Feather-P:
# PR Summary
This PR adds configurable pins for the shared SPI bus:

- `PIN_IMU_SCK`
- `PIN_IMU_MISO`
- `PIN_IMU_MOSI`

The shared SPI bus configuration is integrated into:

- `board-defaults.schema.json`
- `board-defaults.json`
- `scripts/preprocessor.py`
- `DirectSPIInterface`

Chip select remains configured separately for each sensor through `PinInterface`.

## Platform behavior

- On ESP32, configured pins are passed to `SPIClass::begin(sck, miso, mosi)`.
- On ESP8266, the supported pin set is selected through `SPIClass::pins()`.
- If SPI bus pins are not configured, the existing parameterless `SPI.begin()` behavior is preserved.
- ESP8266 hardware CS is disabled after initialization because CS is managed separately for each sensor.

## Testing

Tested locally with:

- ESP32-C3 and LSM6DSV over SPI
- Custom SCK, MISO and MOSI pins
- ESP8266 build compatibility (Not tested on real chips)

## LLM usage

DeepSeek and ChatGPT were used only for debugging assistance and translating commit messages.

Related to #537

### DIFF
--- a/board-defaults.json
+++ b/board-defaults.json
@@ -60,6 +60,11 @@
 						"sda": "4"
 					}
 				],
+				"SPI": {
+					"sck": "14",
+					"miso": "12",
+					"mosi": "13"
+				},
 				"BATTERY": {
 					"type": "BAT_EXTERNAL",
 					"r1": 10,
--- a/board-defaults.json
+++ b/board-defaults.json
@@ -4,22 +4,25 @@
 	"defaults": {
 		"BOARD_SLIMEVR": {
 			"values": {
-				"SENSORS": [
+				"BUSES": [
 					{
 						"protocol": "I2C",
-						"imu": "IMU_AUTO",
-						"int": "16",
-						"rotation": "DEG_270",
+						"sda": "14",
 						"scl": "12",
-						"sda": "14"
-					},
-					{
-						"protocol": "I2C",
-						"imu": "IMU_AUTO",
-						"int": "13",
-						"rotation": "DEG_270",
-						"scl": "12",
-						"sda": "14"
+						"imus": [
+							{
+								"protocol": "I2C",
+								"imu": "IMU_AUTO",
+								"int": "16",
+								"rotation": "DEG_270"
+							},
+							{
+								"protocol": "I2C",
+								"imu": "IMU_AUTO",
+								"int": "13",
+								"rotation": "DEG_270"
+							}
+						]
 					}
 				],
 				"BATTERY": {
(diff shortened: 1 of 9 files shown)


unlogisch04:
With the current board-defaults.json the server crashes when selecting a board different than "BOARD_SLIMEVR_V1_2"

The Flasher generates the Interface over that board-defaults.json

I used a modified board-defaults.json to add the SPI everywhere so the flasher interfaces does not crash for the screenshot.
As you see in the screenshot the SPI settings seem quit a bit out of touch.

So either add it to the SPI_IMU (SCK MISO MOSI)

Best implementation would be that it is BUS -> I2C/SPI -> I2C_IMU/SPI_IMU
But i will not decide here.

I did not test any functions of the firmware.

grafik

Feather-P:
Yeah, I agree. Sorry for not checking how the Server flasher handles the board configuration. I normally flash boards on PlatformIO, completely overlooked the flasher :O
I think the bus configuration could be structured roughly like this:

BUS:
- protocol: I2C or SPI

For I2C:
- SDA
- SCL
- I2C_IMUs
  - IMU type
  - INT pin
  - address
  - rotation
  
For SPI:
- SCK
- MISO
- MOSI
- SPI_IMUs
  - IMU type
  - INT pin
  - CS pin
  - rotation
  
The selected bus protocol would determine which bus fields and IMU schema are displayed, so the protocol would not need to be selected again for every IMU.

Feather-P:
I have implemented the schema and preprocessor changes for BUS -> I2C/SPI -> I2C_IMU/SPI_IMU

But... it seems that some parts of the frontend may need to be adjusted for this new schema.

Everything looks okay when the window is narrower:

image

When I make the window slightly wider:

image

The whole layout gets compressed.

It seems that, at this responsive breakpoint, the Bus group is rendered as a two-column grid, while the nested IMU list is also rendered as a two-column grid. As a result, the whole IMU list only occupies the left cell of the parent grid and is then divided into two columns again inside that cell.

Another problem occurs when trying to add a second Bus. The frontend crashes with:

```text
Uncaught Error: Expected array data but received non-array
    at handleArrayNode (firmware-tool.ts:460:11)
    at handleNode (firmware-tool.ts:654:14)
    at handleDiscriminatedOneOf (firmware-tool.ts:427:12)
    at handleObjectNode (firmware-tool.ts:317:12)
    at handleNode (firmware-tool.ts:650:14)
```

When a new Bus is added, the frontend only initializes:

```json
{
  "protocol": "I2C"
}
```

However, the selected `I2C_BUS` schema also contains an `imus` array. The new object does not initialize this property.

This seems like it could be handled in a separate PR for the Server repository, although it is directly related to the schema changes in this PR.

